Issue #1335, opened 2024-06-28
support much more config file type


kagawagao:
add `.ts`/`.mjs` config file type support for type intelligence and code auto complete

we can consider use [`cosmiconfig`](https://github.com/cosmiconfig/cosmiconfig) for this feature

sorrycc:
I'll close it and reopen it if necessary.

PR #1342, opened 2024-07-01
feat: add much more config file type support

kagawagao:
- use [`cosmiconfig`](https://github.com/cosmiconfig/cosmiconfig) to load common used config file types
- compatible with older `mako.config.json`
- add `defineConfig` helper for config type checking

fixes: #1335

## Summary by CodeRabbit

- **新功能**
  - 使用 `cosmiconfig` 进行配置加载，提高配置处理效率。

- **依赖更新**
  - 添加了 `@swc/helpers` 和 `cosmiconfig` 依赖。
  - 更新了 `less` 和 `less-plugin-resolve` 到最新版本。

### DIFF
--- a/packages/mako/package.json
+++ b/packages/mako/package.json
@@ -20,9 +20,10 @@
   },
   "license": "MIT",
   "dependencies": {
-    "@types/resolve": "^1.20.6",
     "@swc/helpers": "0.5.1",
+    "@types/resolve": "^1.20.6",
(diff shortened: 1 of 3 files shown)


sorrycc:
Thanks for the PR. Configuring support for multiple file types is a complex task because we also read mako.config.json at the Rust layer, and these need to remain consistent. So, I think it's not a right time to do this for now.

PR #1505, opened 2024-08-19
feat: reapply 1499 and fix import meta env

xusd320:
Reapply https://github.com/umijs/mako/pull/1499.
Fix import.meta.env transpiling. The import.meta.env is a static global variable, should not be treated as a defined env.

## Summary by CodeRabbit

- **新特性**
  - 引入了对 `import.meta.env` 的处理，增强了环境变量管理功能。
  - 新增了测试用例，确保 `import.meta.env` 的属性和值正确。
  
- **bug 修复**
  - 精简了环境变量处理逻辑，提高了代码的可维护性和性能。
  
- **文档**
  - 新增了测试环境配置文件，优化了测试流程。

### DIFF
--- a/crates/mako/src/visitors/env_replacer.rs
+++ b/crates/mako/src/visitors/env_replacer.rs
@@ -4,15 +4,11 @@ use std::sync::Arc;
 
 use anyhow::{anyhow, Result};
 use serde_json::Value;
-use swc_core::common::collections::AHashMap;
-use swc_core::common::sync::Lrc;
 use swc_core::common::{Mark, DUMMY_SP};
 use swc_core::ecma::ast::{
     ArrayLit, Bool, ComputedPropName, Expr, ExprOrSpread, Ident, KeyValueProp, Lit, MemberExpr,
-    MemberProp, MetaPropExpr, MetaPropKind, ModuleItem, Null, Number, ObjectLit, Prop, PropName,
-    PropOrSpread, Stmt, Str,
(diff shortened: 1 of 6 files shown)


crates/mako/src/visitors/env_replacer.rs
```diff
@@ -4,172 +4,122 @@ use std::sync::Arc;
...
+                    .split('.')
+                    .rev()
+                    .collect::<Vec<&str>>()
+                    .join(".");
+
+                if let Some(env) = self.get_define_env(&member_visit_path) {
+                    *expr = env
+                }
+            }
+            Expr::Member(MemberExpr {
```
stormslowly, line 95: try case 
```
process.env["A"]
process["env"].A
```
i think those two case can be handled in an `Expr::Member(MemberExpr)` match

crates/mako/src/visitors/import_meta_env_replacer.rs
```diff
@@ -0,0 +1,126 @@
...
+                    }
+                    .into(),
+                });
+            }
+
+            _ => member_expr.visit_mut_children_with(self),
+        }
+    }
+
+    fn visit_mut_expr(&mut self, expr: &mut Expr) {
```
stormslowly, line 62: visit_mut_expr and visit_mut_member_expr are duplicated, maybe just to keep the visist_mut_expr is enough.

but add the below case

```js
let x = () => import.meta.env
```

PR #1554, opened 2024-08-31
perf: group chunks with post order dfs

xusd320:
The previous impl of chunk dependencies by BFS is not right, and in that way, must use Iter::position (compute as O(n)) api to insert depencies before parent. Just change the way by post order DFS, and remove calls of Iter::position.

Why before impl it not right? With below files:

### a.css
```css
@import "b.css";
@import "c.css";

.demo {
 color: red
}
```
### b.css
```css
@import "c.css";

.test {
 color: blue
}
```
### c.css
```css
@import "c.css";

.demo {
 color: yellow
}
```
### index.css
```css
@import "a.css";
@import "b.css";
```

By BFS impl, the ordered modules will be 
```
c.ss
b.css
c.ss
a.css
b.css
index.css
```
And after deduplicating, the modules order will be
```
c.ss
a.css
b.css
index.css
```
 the result css chunk is
```
.demo {
 color: yellow
}
.demo {
 color: red
}
.test {
 color: blue
}
```
Because in index.css, b.css is been imported after a.css, and b.css also import c.css, so c.css will overwrite a.css again. So that this result is not correct.

By post order DFS, the ordered modules will be 
```
c.ss
b.css
c.ss
a.css
c.css
b.css
index.css
```
That result is correct.
Coincidely, in parcel-bundler, these is a same case, https://github.com/parcel-bundler/lightningcss/blob/ad2ec9f10309ac27261a996210023b8e8ad3c793/src/bundler.rs#L1632, also see the same discussion https://github.com/parcel-bundler/parcel/issues/5840.

When building big project, this impl is 8x faster than before:
Before:
image

After:
image

## Summary by CodeRabbit

- **新功能**
	- 改进了模块依赖处理逻辑，增强了模块访问的灵活性和效率。
	- 优化了图结构中的边添加逻辑，提高了代码可读性和准确性。
	- 引入了多个测试脚本，以验证CSS合并功能在不同场景下的正确性。
- **修复**
	- 解决了与 `petgraph` 库相关的邻接模块顺序问题。

### DIFF
--- a/crates/mako/src/generate/chunk_graph.rs
+++ b/crates/mako/src/generate/chunk_graph.rs
@@ -118,7 +118,7 @@ impl ChunkGraph {
             .map(|idx| self.graph[idx].id.clone())
             .collect::<Vec<ChunkId>>();
         // The neighbors ordering is reversed, see https://github.com/petgraph/petgraph/issues/116,
-        // so need to add edges by reversed order
+        // so need to collect by reversed order
         ret.into_iter().rev().collect()
     }
 
--- a/crates/mako/src/generate/chunk_graph.rs
+++ b/crates/mako/src/generate/chunk_graph.rs
@@ -111,11 +111,15 @@ impl ChunkGraph {
 
     pub fn sync_dependencies_chunk(&self, chunk_id: &ChunkId) -> Vec<ChunkId> {
         let idx = self.id_index_map.get(chunk_id).unwrap();
-        self.graph
+        let ret = self
+            .graph
             .neighbors_directed(*idx, Direction::Outgoing)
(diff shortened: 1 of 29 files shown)


stormslowly:
nice shot！ 

rightmost and deepest brings highest priority is so awesome！

stormslowly:
after this pr is  `pub modules: LinkedHashSet<ModuleId>,`   still necessary?

xusd320:
> after this pr is  `pub modules: LinkedHashSet<ModuleId>,`  in Chunk struct  still necessary?

We need a hash_module check, and need iter by insertion order, at this point, there is no difference between LinkedHashSet and IndexSet.

stormslowly (approved):

PR #1582, opened 2024-09-11
fix: bad output when chunk_loading_global containing quotation mark

xusd320:
If chunk_loading_global contains single quote, the output will be broken.

```js
((typeof globalThis !== 'undefined' ? globalThis : self)['makoChunk_application's name'] 
```

## Summary by CodeRabbit

- **新功能**
	- 改进了对 `chunk_loading_global` 属性的处理，增强了生成的 JavaScript 代码的安全性和可读性。

### DIFF
--- a/crates/mako/src/generate/chunk_pot/str_impl.rs
+++ b/crates/mako/src/generate/chunk_pot/str_impl.rs
@@ -112,12 +112,14 @@ pub(super) fn render_normal_js_chunk(
 ) -> Result<ChunkFile> {
     let (content_buf, source_map_buf) = {
         let pot = chunk_pot;
+
+        // to avoid ' or " been included in chunk_loading_global
+        let safe_prop = serde_json::to_string(&context.config.output.chunk_loading_global).unwrap();
+
         let chunk_prefix_code = format!(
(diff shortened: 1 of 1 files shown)


crates/mako/src/generate/chunk_pot/str_impl.rs
```diff
@@ -112,12 +112,16 @@ pub(super) fn render_normal_js_chunk(
 ) -> Result<ChunkFile> {
     let (content_buf, source_map_buf) = {
         let pot = chunk_pot;
+
+        // to avoid ' or " been included in chunk_loading_global
+        let safe_prop =
+            serde_json::from_str::<serde_json::Value>(&context.config.output.chunk_loading_global)
```
stormslowly, line 118: try this
```rust
    let my_string = String::from(r#"Hello, " world!"#);
    let serialized = serde_json::to_string(&my_string).unwrap();
```


https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=c77dda62d9668b6d3b1e30af52180bee

xusd320: nice

stormslowly (approved):

PR #1584, opened 2024-09-12
feat(tree-shaking): optimize import namespace used all exports to partial used of source modules

stormslowly:
## problem

```js
// index.js
import * as math from "./math.js"
math.square(2)
```

```js
// math.js
export function square(){}
export function cube(){}
```
after building, both `square` and `cube` are kept in dist js, but `cube` is not necessary. 

in this case, we can figure out that only  square is used,  so we can optimize that  math.js 's used exports from `ALL_USED` to`Partial(['square'])`

## how to 

if all accesses to the namespace are "explicit", we can say the namespace used and only used these exports  of the source module.

"explicit"  is define as these two types
* `namespace.foo`
* `namespace['foo']`

if there is an no-explicit access in code, we will stop thes optimize; eg.
```js
console.log(namespace); // this implies that all the exports of the namespace will be used.
```

determining  if all accesses to the namespace are "explicit"  is easy.
in intuition, the count of namespace appears in explicit member expr should equal with  the count of  namespace ident appears; but don't forget the `import * as namespace` 
so `if  count(namespace ident appears)  -  count(namespace_member_access ) == 1 ` we can optimize the namespace usage .

## Summary by CodeRabbit

- **新功能**
  - 引入了新的 `optimize_import_namespace` 功能，优化命名空间导入处理。
  - 新增用于验证树摇功能的测试文件，确保构建过程的输出符合预期。
  - 添加了 `needKeep` 和 `needNotKeep` 函数，提供基本数学运算功能。
  - 新增配置文件以优化模块处理，确保构建过程不合并多个模块。
  - 改进了模块摇动过程中的助手函数管理，增强了优化流程的结构化。

- **修复**
  - 移除了 `TransformOutput` 接口，简化了输出转换的处理方式。


Jinbao1001 (approved):

Jinbao1001 (approved):

Jinbao1001:
+1

PR #1720, opened 2024-12-12
fix: inline css config deserialize

xusd320:
Close https://github.com/umijs/mako/issues/1062

Ref: https://github.com/rust-cli/config-rs/issues/461


stormslowly (approved):
magic

PR #1786, opened 2025-03-24
fix: release flow

elrrrrrrr:
> see https://github.com/umijs/mako/actions/runs/14028431328
* fix `--provenance` error
* add `set -e` to exit when error

image

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -1,6 +1,7 @@
 name: utoo-cli-benchmark
 
-on: [pull_request]
+on:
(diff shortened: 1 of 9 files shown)


.github/workflows/bench.yml
```diff
@@ -1,6 +1,7 @@
 name: utoo-cli-benchmark
 
-on: [pull_request]
+on:
+  workflow_dispatch:
```
elrrrrrrr, line 4: 先关闭 bench，需要 20min 才能执行完，后续 build 开启后继续

PR #1789, opened 2025-03-25
feat(bundler): a new bundler arch benchmarked to next-core

xusd320:
1. Add bundler-core
2. Add less-loader supports.


elrrrrrrr (approved):
LGTM

PR #1795, opened 2025-04-07
fix: deps resolve

elrrrrrrr:
1. Added `Display` Trait to print parent chain in verbose mode
2. Use `deno_semver` to replace `semver` to solve the fixed version matching problem
3. Added `validate_deps` fn to parsing exceptions in concurrent scenarios

### DIFF
--- a/crates/cli/src/cmd/deps.rs
+++ b/crates/cli/src/cmd/deps.rs
@@ -1,6 +1,7 @@
 use crate::helper::{package::serialize_tree_to_packages, ruborist::Ruborist};
+use crate::util::logger::log_warning;
+use crate::util::semver;
 use serde_json::json;
-use semver::{Version, VersionReq};
 use std::collections::HashSet;
 use std::fs;
 use std::path::PathBuf;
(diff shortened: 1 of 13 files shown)


crates/cli/src/cmd/deps.rs
```diff
@@ -80,3 +83,86 @@ pub async fn build_workspace() -> std::io::Result<()> {
...
+fn validate_deps() -> std::io::Result<()> {
+    let path = PathBuf::from(".");
+    let lock_path = path.join("package-lock.json");
+
+    let lock_content = fs::read_to_string(lock_path)?;
+    let lock_file: serde_json::Value = serde_json::from_str(&lock_content)?;
+
+    if let Some(packages) = lock_file.get("packages").and_then(|p| p.as_object()) {
+        for (pkg_path, pkg_info) in packages {
+            let dep_types = [
```
xusd320, line 96: 为啥不 const ？

elrrrrrrr: 我改下🙏🏻

xusd320 (approved):

PR #1796, opened 2025-04-07
feat(bundler): support umd project (PART2)

xusd320:
1. impl LibraryChunkingContext
2. add dynamic_import_to_require transform for libary

More todos for library:
1. inline css
2. output filename
3. umd root
4. umd export


elrrrrrrr:
😄 cargo fmt

elrrrrrrr (approved):

PR #1811, opened 2025-04-15
fix: shibang

elrrrrrrr:
1. fix shibang space


stormslowly:
@elrrrrrrr it's `shebang`; if you pronuance `shibang` in pinyin, ...

PR #1818, opened 2025-04-18
fix: less,sass,postcss loader panic

xusd320:
See: https://github.com/umijs/mako/blob/5f68c35a00688b7ca466948c0ad7a9a39e0acf6e/packages/mako/src/index.ts#L230,
"this" (the class instance) is lost.


PeachScript (approved):
工具人 +1

PR #1866, opened 2025-04-29
feat: global install

elrrrrrrr:
* 🌍 Added `--global` option for install
* 🔄 The auto update logic uses `utoo` to bootstrap, removing `npm` related command
* ⏳ The new version will be used the next time it is called

image


killagu (approved):
LGTM

PR #1874, opened 2025-05-06
fix: save prefix

elrrrrrrr:
> Implement savePrefix Logic with ^ Default

* 🔭 Automatically adds ^ prefix when saving dependencies **only if**:
    * No custom spec is provided (e.g., utoo install package@1.2.3 won't be modified)
    * The version isn't already prefixed (e.g., ~1.2.3, ^1.2.3 remain unchanged)


killagu (approved):
LGTM

PR #1890, opened 2025-05-13
refactor(bundler): support write all entrypoints to dist, fix hmr

xusd320:
As titled.


fireairforce (commented):
another question, what's the concept of `entrypoint` here, just like webpack's https://webpack.js.org/concepts/entry-points/ ?

xusd320:
> https://webpack.js.org/concepts/entry-points/

Yeah, they are very simular.

xusd320:
> another question, what's the concept of `entrypoint` here, just like webpack's https://webpack.js.org/concepts/entry-points/ ?

Yeah, they are very simular.

fireairforce (approved):

PR #1900, opened 2025-05-20
feat(bundler): support app project(only csr)

xusd320:
To support app project bundling basically. Now we have to implement by some hacking way.

## Why the hacking pactch?
Formerly, we thought that we can use every crates under the next-core as bundler infrastructure, now I found that these is some gap between our design and next.js. The hacking is for features of output.filename and ouput.chunkFilename, they are not supported by turbopack officially.
image

## How we patch?
Use a forked repo https://github.com/umijs/next.js, patch by every stable version of next.js, the patched branch will be named by the rule of:
```js
`utoo-patch-v${stableNextjsVersion}`
```
We need to comment about the patch reason. We can see the diff by this form of compare https://github.com/vercel/next.js/compare/v15.3.2...umijs:next.js:utoo-patch-v15.3.2


packages/bundler/src/project.ts
```diff
@@ -261,6 +261,7 @@ export function projectFactory() {
       type NapiEndpoint = { __napiType: "Endpoint" };
 
       type NapiEntrypoints = {
+        apps?: NapiEndpoint[];
```
fireairforce, line 264: we support a config apps to support bundler for application, can u please help to add a example under `examples/*` so we can debug the difference between with `libraries` and `apps`

examples/with-antd/project_options.json
```diff
@@ -4,14 +4,17 @@
   "config": {
     "entry": [
       {
-        "import": "./src/index.tsx",
-        "library": {
```
fireairforce, line 8: the `apps` config is default here?

xusd320: No, library diffs from app only at the point at whether there is a library config in entry options, see https://github.com/webpack/webpack/blob/main/declarations/WebpackOptions.d.ts#L1189, this is an optional config, when it is set, it's a library, when it's not, it's an app.That is very clear, no need to explain be example.

xusd320: We filter the apps and librarys like this
image

fireairforce: get~

fireairforce (approved):

PR #1903, opened 2025-05-22
feat: validate lock when reusing

elrrrrrrr:
> Add validation logic before loading local package-lock.json to resolve inconsistencies caused by manual edits to package.json

* 🚚 Moved `validate_deps` to helper implementation  
* 🆕 Added util json methods to standardize loading of package.json and package-lock.json

image


crates/cli/src/helper/lock.rs
```diff
@@ -236,6 +245,167 @@ pub fn path_to_pkg_name(path_str: &str) -> Option<&str> {
...
+                    for (dep_name, req_version) in dependencies {
+                        let req_version_str = req_version.as_str().unwrap_or_default();
+
+                        // Collect parent chain information
+                        let mut parent_chain = Vec::new();
+                        let mut current_path = if pkg_path.is_empty() {
+                            String::new()
+                        } else {
+                            pkg_path.to_string()
+                        };
```
fireairforce, line 282: How about the following code:

```suggestion
 let mut current_path = String::from(pkg_path);
```

crates/cli/src/helper/lock.rs
```diff
@@ -236,6 +245,167 @@ pub fn path_to_pkg_name(path_str: &str) -> Option<&str> {
...
+                        } else {
+                            req_version_str.to_string()
+                        };
+
+                        // find the actual version of the dependency
+                        let mut current_path = if pkg_path.is_empty() {
+                            String::new()
+                        } else {
+                            pkg_path.to_string()
+                        };
```
fireairforce, line 324: the same as follows

PR #1936, opened 2025-06-04
chore: rename utoo-bundler to utoo-pack

xusd320:
As title.


xusd320:
> it seems conflict with next branch

Resolved.

PR #1947, opened 2025-06-10
fix(pack-cli): hmr entry ident

xusd320:
Should not modify the entry request with `name` query, this will break the hmr subscription.


elrrrrrrr (approved):
LGTM

PR #2058, opened 2025-07-17
chore: bump biome to v2.1.2 to avoid memory leak

fireairforce:
## Summary

Ref: https://github.com/biomejs/biome/issues/6784#issuecomment-3084169206

## Test Plan

No


xusd320 (approved):

biome.json
```diff
@@ -1,5 +1,5 @@
 {
-  "$schema": "https://biomejs.dev/schemas/2.0.5/schema.json",
+  "$schema": "https://biomejs.dev/schemas/2.1.2/schema.json",
```
afc163, line 2: https://github.com/ant-design/ant-design/blob/7083e3db7c0cb43259bff34878e59ba8dbdcf180/biome.json#L2

PR #2061, opened 2025-07-18
feat(pack): let internal dependencies of @utoo/pack be pinned

xusd320:
## Summary

Just let dependencies `@utoo/pack` won't be resolved out of `@utoo/pack`, this is useful for special npm manager like `pnpm` which has different behavior of `hositing`, or frameworks like `umi` which uses `@utoo/pack` as a second level dependency.

## Test Plan

webpack-loaders test added.


fireairforce (dismissed):
这个合完我发个版测一下

fireairforce (approved):
lgtm

PR #2082, opened 2025-07-28
fix(pack): module.rules.loaders support relative path

fireairforce:
## Summary

closes: #2081 

This pr will make custom webpack loader can work fine under utoo-pack.

Next.js(Turbopack) has the same question: https://github.com/vercel/next.js/issues/82106 , and i'm not sure this is design problem or others, i fix this problem by generating an absoulte path for loader path.

Maybe the better way to fix this problem maybe just change the loader resolver methond: https://github.com/vercel/next.js/blob/canary/turbopack/crates/turbopack-node/js/src/transforms/webpack-loaders.ts#L431

This will  affect the logic of next.js and turbopack, so i just change the logic of utoo-pack-core to support this case.

utoo-pack need to support this because our config is JSON, we can't evalute the absolute path in the config file.

## Test Plan

I add test case under `pack-tests` and i test the devServer by `examples/react`:

image


crates/pack-core/src/config.rs
```diff
@@ -777,7 +778,12 @@ impl Config {
     }
 
     #[turbo_tasks::function]
-    pub fn webpack_rules(&self, active_conditions: Vec<RcStr>) -> Vc<OptionWebpackRules> {
+    pub fn webpack_rules(
+        &self,
+        active_conditions: Vec<RcStr>,
+        project_dir: FileSystemPath,
+    ) -> Vc<OptionWebpackRules> {
+        dbg!(&project_dir);
```
xusd320, line 786: del this

crates/pack-core/src/config.rs
```diff
@@ -787,16 +793,47 @@ impl Config {
...
+            fn transform_loaders(
+                loaders: &[LoaderItem],
+                project_dir: &FileSystemPath,
+            ) -> ResolvedVc<WebpackLoaderItems> {
+                // issue: https://github.com/umijs/mako/issues/2081
+                // issue: https://github.com/vercel/next.js/issues/82106
+                fn resolve_loader_path(loader_name: &str, project_dir: &FileSystemPath) -> RcStr {
+                    if loader_name.starts_with("./") || loader_name.starts_with("../") {
+                        // This is a relative path with explicit prefix, convert to absolute path
+                        let cwd = std::env::current_dir().unwrap_or_default();
```
xusd320, line 805: 抽一个 util 放到外面去好了，这个 fn webpack_rules 代码量已经很大了，不要原地增加复杂性

xusd320 (changes requested):

crates/pack-core/src/config.rs
fireairforce: ok

crates/pack-core/src/config.rs
fireairforce: done

crates/pack-core/src/config.rs
fireairforce: done

xusd320 (approved):

PR #2251, opened 2025-10-16
fix(pm): link path

elrrrrrrr:
* 🐞 Fixed incorrect global directory during ut linking
* 💃🏻 Print current package name during ut linking
* 🚚 Removed synchronous APIs from util/linker.rs

### DIFF
--- a/crates/pm/src/service/script.rs
+++ b/crates/pm/src/service/script.rs
@@ -284,30 +284,6 @@ mod tests {
     use tempfile::tempdir;
 
     #[tokio::test]
-    async fn test_execute_custom_script_failure() {
-        let temp_dir = tempdir().unwrap();
-        let package = PackageInfo {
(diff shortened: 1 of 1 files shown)


crates/pm/src/cmd/link.rs
```diff
@@ -7,7 +7,7 @@ use crate::util::logger::log_verbose;
 use anyhow::{Context, Result};
 
 /// Link current package to global (equivalent to npm link without args)
-pub async fn link_current_to_global(prefix: Option<&str>) -> Result<()> {
+pub async fn link_current_to_global(prefix: Option<&str>) -> Result<String> {
```
elrrrrrrr: link_current_to_global 需要返回 package_info.name ，因为 package_info 是内部的局部变量，无法直接返回 as_str(), @xusd320 看看有没有好办法 😯

xusd320 (approved):

PR #2254, opened 2025-10-17
refactor(pm): tracing

elrrrrrrr:
* Replaced logging implementation with unified tracing framework. Key features:
  1. ✅ Supports `--verbose` flag for unified log output
  2. ✅ Added atty check to only show colored output in `TTY` environments
  3 .✅ Better performance compared to `mutex` based solutions
  4. ✅ Automatic error logging to disk
* Removed redundant map_err calls, using context and with_context for stack traces instead
* Centralized log persistence logic in `main.rs`

### DIFF
--- a/crates/pm/src/util/logger.rs
+++ b/crates/pm/src/util/logger.rs
@@ -24,10 +24,10 @@ pub static PROGRESS_BAR: Lazy<ProgressBar> = Lazy::new(|| {
 
 // Global state for tracing
 static LOG_FILE_PATH: OnceCell<PathBuf> = OnceCell::new();
-static _LOG_GUARD: OnceCell<WorkerGuard> = OnceCell::new();
 
 /// Initialize tracing subscriber with console and file output
-pub fn init_tracing(verbose: bool) -> Result<PathBuf> {
+/// Returns (log_path, guard) - the guard must be kept alive for the duration of the program
+pub fn init_tracing(verbose: bool) -> Result<(PathBuf, WorkerGuard)> {
     // 1. Build environment filters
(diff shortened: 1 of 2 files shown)


crates/pm/src/util/logger.rs
```diff
@@ -21,6 +22,71 @@ pub static PROGRESS_BAR: Lazy<ProgressBar> = Lazy::new(|| {
     pb
 });
 
+// Global state for tracing
+static LOG_FILE_PATH: OnceCell<PathBuf> = OnceCell::new();
+static _LOG_GUARD: OnceCell<WorkerGuard> = OnceCell::new();
+
+/// Initialize tracing subscriber with console and file output
+pub fn init_tracing(verbose: bool) -> Result<PathBuf> {
```
xusd320: 为啥不用精准的 Filter? 可以过滤到 mod 级别，方便查问题。
` RUST_LOG=reqwest=debug,pi=info  ~/Codes/github/pi/target/release/pi`
默认给 utoo = "info" 就好了。

crates/pm/src/util/logger.rs
```diff
@@ -21,6 +22,71 @@ pub static PROGRESS_BAR: Lazy<ProgressBar> = Lazy::new(|| {
     pb
 });
 
+// Global state for tracing
+static LOG_FILE_PATH: OnceCell<PathBuf> = OnceCell::new();
+static _LOG_GUARD: OnceCell<WorkerGuard> = OnceCell::new();
```
xusd320: 这没必要 static 吧....

crates/pm/src/util/logger.rs
elrrrrrrr: Image

PR #2355, opened 2025-11-29
fix(pm): binary permission

elrrrrrrr:
1. Fixed installation failure for non-UTF-8 binary files
2. add `check_and_add_shebang` util fn，cc @stormslowly

### DIFF
--- a/crates/pm/src/service/script.rs
+++ b/crates/pm/src/service/script.rs
@@ -535,11 +535,7 @@ mod tests {
 
         let result = ScriptService::check_and_add_shebang(&text_file).await;
         assert!(result.is_ok(), "Should succeed for text file");
-        assert_eq!(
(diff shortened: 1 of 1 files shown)


crates/pm/src/service/script.rs
```diff
@@ -192,6 +194,38 @@ impl ScriptService {
...
+        let header = std::str::from_utf8(&buffer)
+            .map_err(|_| anyhow::anyhow!("File is not valid UTF-8, likely a binary file"))?;
+
+        // Check if already has shebang
+        if header.starts_with("#!") {
+            return Ok(false);
+        }
+
+        // Need to add shebang - read entire file now
+        drop(file);
```
xusd320: 尽量少用显式 drop, 用 block 自动 RAII

PR #2391, opened 2025-12-10
fix(wasm): sw proxy

xusd320:
fix the error of "Failed to execute 'respondWith' on 'FetchEvent': The event handler is already finished"


fireairforce (dismissed):

packages/utoo-web/src/serviceWorker.ts
```diff
@@ -34,21 +34,25 @@ self.addEventListener("message", (event) => {
...
+  event.respondWith(
+    (async () => {
+      await _promise;
+      let { url: url_str, referrer } = event.request;
+      let url = new URL(url_str);
+      if (
+        url.pathname.startsWith(_serviceWorkerScope) ||
+        (referrer && new URL(referrer).pathname.startsWith(_serviceWorkerScope))
+      ) {
+        const relateivePathToCwd =
```
kuitos:
```suggestion
        const relativePathToCwd =
```

typo

PR #2567, opened 2026-02-03
refactor(pm): simplify Linux cloner with sync hardlink strategy

elrrrrrrr:
* The previous implementation tried `FICLONE ioctl`, then `copy_file_range`,
then `regular copy` - each with separate spawn_blocking calls.

* This simplification uses hardlinks (which are fast and share disk space)
with copy fallback, all in a single `blocking task`.

### DIFF
--- a/crates/pm/src/util/cloner.rs
+++ b/crates/pm/src/util/cloner.rs
@@ -45,10 +45,10 @@ mod linux_clone {
     // Recursive directory copy with hardlink strategy (sync version)
     fn clone_dir_sync(src: &Path, dst: &Path, use_copy: bool) -> io::Result<()> {
         // Create destination directory
-        if let Err(e) = fs::create_dir(dst) {
-            if e.kind() != io::ErrorKind::AlreadyExists {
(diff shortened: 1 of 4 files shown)


crates/pm/src/util/cloner.rs
```diff
@@ -17,217 +17,89 @@ use std::os::unix::ffi::OsStrExt;
...
-            let ret = unsafe { libc::ioctl(dst_fd, libc::FICLONE, src_fd) };
-
-            if ret < 0 {
-                let err = std::io::Error::last_os_error();
-                Err(err)
-            } else {
-                Ok(())
+    // Check if the package has install scripts (sync version)
+    fn has_install_script_sync(src: &Path) -> bool {
+        if let Some(parent) = src.parent() {
```
xusd320: /gemini 进一步简化下 https://doc.rust-lang.org/std/option/enum.Option.html#method.is_some_and

xusd320 (approved):

fireairforce (approved):

PR #2595, opened 2026-02-12
feat(pack): support config.react.importSource

fireairforce:


crates/pack-core/src/transform_options.rs
```diff
@@ -117,52 +114,32 @@ pub async fn get_decorators_transform_options(
...
-            None
-        },
-        runtime: react_config.runtime.as_ref().map(|runtime| match runtime {
-            ReactRuntime::Automatic => "automatic".into(),
-            ReactRuntime::Classic => "classic".into(),
-        }),
+        import_source: react_config
+            .import_source
+            .clone()
+            .or(if is_emotion_enabled {
```
xusd320: emotion 可以删了

fireairforce: done 了

PR #2599, opened 2026-02-12
feat(pm): utoo ping

elrrrrrrr:
* 🧶 Add new `utoo ping` command - similar to `npm ping`, it displays registry latency and returns whether semver is supported.
* ♻️ Refactor existing config.rs module
  1. Configuration read/write operations through `config_file.rs`
  2. Runtime configuration exposed through `user_config.rs`
* 🤖 Auto-detect registry semver support - No longer hardcoded, persistent results stored in `config.toml`

image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -885,6 +885,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "clap_complete"
+version = "4.5.66"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c757a3b7e39161a4e56f9365141ada2a6c915a8622c408ab6bb4b5d047371031"
+dependencies = [
+ "clap",
+]
+
+[[package]]
 name = "clap_derive"
 version = "4.5.47"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -9703,6 +9712,7 @@ dependencies = [
  "bytes",
  "chrono",
  "clap",
+ "clap_complete",
  "colored 2.2.0",
  "dashmap 6.1.0",
  "deno_semver",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8857,6 +8857,7 @@ dependencies = [
  "serde_json",
  "shrink-to-fit",
  "smallvec",
+ "thiserror 1.0.69",
  "tokio",
  "tokio-util",
  "tracing",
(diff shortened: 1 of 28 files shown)


killagu (commented):
## Code Review

### Overall

The refactoring is well-structured and the new ping command is a useful addition. The `run.rs` test improvements (removing `set_current_dir`) are a nice side benefit. A few issues worth addressing:

---

### 1. Race condition in semver detection (High)

`set_registry` → `detect_supports_semver` loads config, probes the network, then writes back. There's no file locking — concurrent `utoo` processes could race on the config file and lose entries.

```rust
// user_config.rs
let mut entries = config.get_array("supports-semver").cloned().unwrap_or_default();
entries.push(new_entry);
config.set_array("supports-semver", entries, true)
```

Consider using file locking (e.g., `fs2::FileExt`) when writing the global config, or at minimum document this as a known limitation.

### 2. Unbounded growth of `supports-semver` array (Medium)

Every unique registry URL gets appended but entries are never deduplicated or pruned. If a user switches registries frequently, this array grows forever. Consider:
- Deduplicating before appending (check if the URL or `!URL` already exists and update in-place)
- Capping the array size

### 3. `ping` returns `Ok(())` even on failure (Medium)

```rust
// ping.rs
} else {
    println!("FAIL registry did not respond ({}ms)", result.latency_ms);
}
Ok(())
```

A failed ping still returns `Ok(())`, so `utoo ping` exits with code 0 even when the registry is unreachable. Consider returning an error on failure so scripts/CI can rely on the exit code.

### 4. `OnceLock` for `SUPPORTS_SEMVER` — set-once but registry can change (Medium)

`SUPPORTS_SEMVER` is `OnceLock<bool>`, so once set it cannot be updated. If `set_registry` is called multiple times (e.g. in tests), only the first semver result sticks:

```rust
let _ = SUPPORTS_SEMVER.set(supports);  // silently fails on second call
```

Consistent with other config values using the same pattern, but worth noting.

### 5. `ping` command creates a redundant `reqwest::Client` (Low)

In `cmd/ping.rs`, a new client is built with 10s timeout, then `detect_supports_semver` creates another client with 5s timeout internally. Consider reusing the client or making the timeout consistent.

### 6. `list_arrays` output uses `Debug` format (Nit)

```rust
for (key, values) in config.list_arrays() {
    println!("{key} = {:?}", values);
}
```

`{:?}` will print Rust-style `["url1", "url2"]`. Consider a more user-friendly format for CLI output.

---

Good work overall 👍 The main actionable items are #2 (dedup) and #3 (exit code).

killagu (approved):
LGTM

PR #2614, opened 2026-02-24
fix(multi-entries-heavy): correct devdependencies -> devDependencies

### LINKED ISSUE #2613: 🐛 [Build] turbo build with wrong dependencies
### What happened?

Image

Turbo fails to resolve dependencies of `examples/multi-entry-heavy`, causing it to be built concurrently with `@utoo/pack-shared`. This happens because `@utoo/pack` has not yet been built in the monorepo pipeline.

As a result, running `ut build` from the root without a prior successful build will fail.

Image

### Expected result

_No response_

yuzheng14:
So npm includes pack-cli in lockfile and turbo respects build order.

Close #2613.

### DIFF
--- a/examples/multi-entries-heavy/package.json
+++ b/examples/multi-entries-heavy/package.json
@@ -15,7 +15,7 @@
     "react-dom": "18.2.0",
     "three": "^0.153.0"
   },
-  "devdependencies": {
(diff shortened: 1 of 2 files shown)


fireairforce (approved):
lgtm

PR #2645, opened 2026-02-27
fix(pack): library output stats & chunk path

xusd320:
1. fix the bad magic mangle for symbol `__TURBOPACK__`
2. generate `stats.json` for `library` output
3. keep the `output.filename` dir prefix, emit chunks in that dir

### DIFF
--- a/crates/pack-core/src/library/runtime/runtime_code.rs
+++ b/crates/pack-core/src/library/runtime/runtime_code.rs
@@ -48,7 +48,7 @@ pub async fn get_library_runtime_code(
     writedoc!(
         code,
         r#"            
-            if (!Array.isArray(__TURBOPACK__)) {{
+            if (!Array.isArray(__UTOOPACK__)) {{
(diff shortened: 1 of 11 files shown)


crates/pack-core/src/library/runtime/runtime_code.rs
```diff
@@ -93,8 +93,8 @@ pub async fn get_library_runtime_code(
     writedoc!(
         code,
         r#"
-            const chunksToRegister = __TURBOPACK__;
-            __TURBOPACK__ = {{ push: registerChunk }};
+            const chunksToRegister = __UTOOPACK__;
+            __UTOOPACK__ = {{ push: registerChunk }};
```
fireairforce, line 97: 这个要更新快照

xusd320: Done

fireairforce (approved):

PR #2648, opened 2026-02-28
fix(dev): simplify publicPath check in initialize function

xc1427:
## Summary

Due to this change https://github.com/umijs/umi/pull/13257, the `runtime` check is no longer needed and is considered as dead code. We here simply remove it.

## Test Plan

It is eliminating a "condition branch" so current test suite is enough.


fireairforce (changes requested):
稍等，我先用的你 example 复现跑一下，这里 utoopack 的行为在 umi 框架应该跟其他的 bundler 都是一致的

xc1427:
closed in favor of https://github.com/umijs/umi/pull/13258

PR #2684, opened 2026-03-12
fix(pack): compatWebpack treeshaking config

fireairforce:
## Summary

Default value of webpack's unsedexports is true: https://webpack.js.org/configuration/optimization/#optimizationusedexports

It means even if it's undefined(user don't config), it will enable.

`!!undefined === false` in javascript, so in the default case utoopack will disable tree shaking(but webpack will enable).

### DIFF
--- a/packages/pack-shared/src/webpackCompat.ts
+++ b/packages/pack-shared/src/webpackCompat.ts
@@ -697,7 +697,7 @@ function compatOptimization(
     noMangling: webpackOptimization.mangleExports === false,
     minify: minimize,
     concatenateModules,
-    treeShaking: enableTreeShaking,
+    treeShaking: false,
     removeUnusedExports: enableTreeShaking,
     removeUnusedImports: enableTreeShaking,
   };
--- a/packages/pack-shared/src/webpackCompat.ts
+++ b/packages/pack-shared/src/webpackCompat.ts
@@ -686,7 +686,7 @@ function compatOptimization(
   }
   const { moduleIds, minimize, concatenateModules, usedExports } =
     webpackOptimization;
-  const enableTreeShaking = usedExports !== false;
+  const enableWebpackUsedExports = usedExports !== false;
   return {
     moduleIds:
       moduleIds === "named"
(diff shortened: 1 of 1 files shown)


packages/pack-shared/src/webpackCompat.ts
```diff
@@ -696,9 +697,9 @@ function compatOptimization(
     noMangling: webpackOptimization.mangleExports === false,
     minify: minimize,
     concatenateModules,
-    treeShaking: !!usedExports,
```
xusd320, line 699: tree shaking 现在不能开，开了会用 Fragments 模式，会挂掉

fireairforce: 设置成 false 了

packages/pack-shared/src/webpackCompat.ts
```diff
@@ -686,6 +686,7 @@ function compatOptimization(
   }
   const { moduleIds, minimize, concatenateModules, usedExports } =
     webpackOptimization;
+  const enableTreeShaking = usedExports !== false;
```
xusd320:
```suggestion
  const usedExports = usedExports !== false;
```

packages/pack-shared/src/webpackCompat.ts
```diff
@@ -696,9 +697,9 @@ function compatOptimization(
     noMangling: webpackOptimization.mangleExports === false,
     minify: minimize,
     concatenateModules,
-    treeShaking: !!usedExports,
-    removeUnusedExports: !!usedExports,
-    removeUnusedImports: !!usedExports,
+    treeShaking: false,
+    removeUnusedExports: enableTreeShaking,
```
xusd320:
```suggestion
    removeUnusedExports: usedExports,
```

packages/pack-shared/src/webpackCompat.ts
```diff
@@ -696,9 +697,9 @@ function compatOptimization(
...
+    removeUnusedImports: enableTreeShaking,
```
xusd320:
```suggestion
    removeUnusedImports: usedExports,
```

packages/pack-shared/src/webpackCompat.ts
fireairforce: 不能叫这个，会和上面取值冲突

xusd320 (approved):

PR #2702, opened 2026-03-16
fix(pm): Windows support — normalize OS and use sh for postinstall

elrrrrrrr:
## Summary

Fix utoo on Windows by addressing two root causes:

1. **`normalize_os` missing win32/windows mapping** — Rust's `std::env::consts::OS` returns `"windows"` but npm packages use `"win32"` in their `os` field. Without normalization, platform-specific optional dependencies like `@rolldown/binding-win32-x64-msvc` are incorrectly skipped on Windows.

2. **postinstall uses bash which may not be in PATH on Windows** — Change from `bash` to `sh` (POSIX compatible). Git for Windows provides `sh.exe` in PATH. Also convert `[[ ]]` bash-only syntax to POSIX `[ ]` in all templates.

## Changes

- `crates/ruborist/src/model/compatibility.rs`: Add `"win32" | "windows" => "win32"` to `normalize_os`
- `vendor/templates/*.template`: `#!/bin/bash` → `#!/bin/sh`, `[[ ]]` → `[ ]`
- `vendor/templates/utoo.package.json.template`: `bash ./postinstall.sh` → `sh ./postinstall.sh`
- `vendor/scripts/npm-utoo.sh`: placeholder shebangs `#!/bin/bash` → `#!/bin/sh`
- `e2e/utoo-pm.sh`, `e2e/utoo-pm.ps1`: Add optional deps test case (rolldown binding)

## Test plan

- [x] `cargo test -p utoo-ruborist -- compatibility` passes
- [ ] E2E: rolldown binding installed on Windows/macOS/Linux


killagu (approved):
LGTM

PR #2727, opened 2026-03-24
fix(pm): smarter workspace devDep cycle handling

elrrrrrrr:
## Summary

- Workspace topology computation now **only drops dev edges within the same SCC** (strongly connected component) when a cycle is detected, instead of dropping all dev edges. This preserves ordering constraints from non-cyclic devDependencies.
- Simplified `compute_topological_layers` API: removed `Node`/`Edge` wrapper types, now accepts `&[String]` + `&[(&str, &str)]` directly. Moved edge struct to `workspace.rs` as `WorkspaceEdge`.
- Extracted `find_cycle_groups` (SCC-based) to reuse cycle detection logic.

## Test plan

- [x] Unit tests in `deps.rs` (16 tests including `find_cycle_groups`)
- [x] Unit tests in `workspace.rs` (4 tests: devDep cycle fallback, non-cyclic devDep ordering, mixed prod+dev cycle, JSON output)
- [x] E2e fixture `e2e/pm/workspace-cycle/` with `lib-a ↔ lib-b` dev cycle, verified via `ut deps --workspace-only`
- [x] CI (`cargo test`, `cargo clippy`, e2e)


crates/pm/src/helper/deps.rs
```diff
@@ -1,118 +1,87 @@
...
-            graph.add_edge(from_idx, to_idx, ());
-        }
-    }
+/// Each returned `Vec<String>` is one cycle group. Returns empty when acyclic.
+pub fn find_cycle_groups(node_list: &[String], edges: &[(&str, &str)]) -> Vec<Vec<String>> {
+    let graph = build_graph(node_list, edges);
+    // kosaraju_scc returns all strongly connected components (SCCs) — maximal
+    // sets of nodes where every node is reachable from every other. An SCC
+    // with >1 node (or a single self-loop) is a cycle.
+    kosaraju_scc(&graph)
```
xusd320, line 35: 这么个高级的算法都用上了

xusd320 (approved):

PR #2733, opened 2026-03-24
feat(pm): add `--from pnpm` migration for install command

elrrrrrrr:
## Summary

- Add `utoo install --from pnpm` to migrate pnpm projects to utoo
- Parses `pnpm-workspace.yaml` and converts config:
  - `packages` → `package.json` `workspaces` field
  - `overrides` → `package.json` `overrides` field
  - `catalog`/`catalogs` → `.utoo.toml`
- Zero core logic changes — migration writes files, existing install picks them up natively
- Added `FromPm` enum with `clap::ValueEnum` for type-safe `--from` flag

## Test plan

- [x] Unit tests: 5 new tests covering YAML parsing, package.json update, .utoo.toml generation, error handling
- [x] All 234 existing tests pass
- [x] cargo fmt + clippy clean
- [x] E2E test: `eggjs/egg` (next branch) pnpm migration + install

### DIFF
--- a/crates/pm/src/main.rs
+++ b/crates/pm/src/main.rs
@@ -44,6 +44,8 @@ use crate::constants::cmd::{
 use crate::constants::{APP_ABOUT, APP_NAME, APP_VERSION};
 use crate::helper::workspace::init_project_root;
 
+use crate::helper::migrate::{FromPm, migrate_from_pnpm};
+
 fn detect_shell_from_env() -> Option<clap_complete::Shell> {
     // Most common on Unix-like systems.
     let shell_path = std::env::var("SHELL").ok()?;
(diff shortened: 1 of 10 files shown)


zombieJ:
@codex review

crates/pm/src/main.rs
```diff
@@ -399,6 +405,17 @@ async fn async_main() -> Result<()> {
...
     );
 
+    // Run --from migration early, before config is cached by init_registry
+    if let Some(Commands::Install {
+        from: Some(FromPm::Pnpm),
+        ..
+    }) = &cli.command
+    {
+        let cwd = std::env::current_dir()?;
+        let root_path = init_project_root(&cwd).await?;
```
kuitos, line 415: Calling `init_project_root()` here assumes the repo root is already discoverable, but before migration a pnpm monorepo is only rooted by `pnpm-workspace.yaml`. If the command is run from `packages/foo`, `find_root_path()` still stops at the nearest `package.json` because workspace-root detection only recognizes `package.json.workspaces`, so this ends up looking for `packages/foo/pnpm-workspace.yaml` and fails. Could we resolve the migration root by searching upward for `pnpm-workspace.yaml` first, or teach workspace discovery to recognize pnpm roots in this path?

elrrrrrrr: Image

init_project_rooot 会自动寻址到根路径

fireairforce (approved):

PR #2764, opened 2026-04-07
fix(pack): support merge nested define config

### LINKED ISSUE #2763: utoopack does not correctly handle Umi's define configuration for module-level constants
## Bug Description

Umi's `define` configuration is not correctly handling module-level constants in utoopack.

## Reproduction Steps

1. Define in `config/config.ts`:
```typescript
define: {
  'process.env.CI': process.env.CI,
},
```

2. Use in `src/app.tsx`:
```typescript
const isDev = process.env.NODE_ENV === 'development';
const isDevOrTest = isDev || process.env.CI;
```

3. Build with utoopack: `CI=true npm run build`

4. Check the compiled output - the `isDevOrTest` variable is declared but never assigned a value

## Expected Behavior

When `CI=true`, `process.env.CI` should be replaced with `"true"` (or `true`), making `isDevOrTest` a truthy value, and the SettingDrawer should be displayed.

## Actual Behavior

After building with utoopack, the code corresponding to `isDevOrTest` is declared but never assigned (always `undefined`), causing the SettingDrawer to never display.

Compiled output shows:
```javascript
let eG, eJ = "/user/login";  // eG is declared but never assigned
// ...later...
{children: [n, eG && (0, s.jsx)(l.SettingDrawer, ...)]}  // eG is always undefined
```

## Comparison with mako

When using `mako` instead of `utoopack`, the environment variable is correctly handled and the SettingDrawer displays as expected.

With mako + `CI=true`: the condition is properly folded and SettingDrawer is included.
With mako + no CI: the condition is folded to false and SettingDrawer is tree-shaken.

## Environment

- umi version: 4.6.x
- utoopack version: latest

## Related

- PR: https://github.com/ant-design/ant-design-pro/pull/11644

fireairforce:
## Summary

fix: https://github.com/utooland/utoo/issues/2763

This case need to fix in upstream too.

Before:

image

After:

image

## Test Plan

add snapshot test case

### DIFF
--- a/crates/pack-core/src/shared/contexts.rs
+++ b/crates/pack-core/src/shared/contexts.rs
@@ -81,10 +81,7 @@ fn defines_from_ref(define_env: &FxIndexMap<RcStr, RcStr>) -> CompileTimeDefines
     for (key, value) in define_entries {
         for prefix_len in 1..key.len() {
             let (prefix, suffix) = key.split_at(prefix_len);
-            if let Some((_, parent)) = defines
-                .iter_mut()
-                .find(|(candidate, _)| candidate.as_slice() == prefix)
-            {
+            if let Some(parent) = defines.get_mut(prefix) {
                 let _ = merge_nested_define(parent, suffix, value.clone());
             }
         }
--- a/crates/pack-core/src/shared/contexts.rs
+++ b/crates/pack-core/src/shared/contexts.rs
@@ -14,23 +14,80 @@ use turbopack_core::{
 
 use crate::config::{ProviderConfig, ProviderConfigValue};
 
+fn parse_define_value(value: &RcStr) -> CompileTimeDefineValue {
+    match serde_json::Value::from_str(value) {
+        Ok(value) => value.into(),
+        Err(_) => CompileTimeDefineValue::Evaluate(value.clone()),
+    }
+}
+
+fn merge_nested_define(
+    current: &mut CompileTimeDefineValue,
+    path: &[DefinableNameSegment],
+    value: CompileTimeDefineValue,
+) -> bool {
+    if path.is_empty() {
+        *current = value;
+        return true;
+    }
+
+    match current {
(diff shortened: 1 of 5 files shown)


crates/pack-core/src/shared/contexts.rs
```diff
@@ -14,23 +14,80 @@ use turbopack_core::{
...
+        .map(|(key, value)| (key.clone(), value.clone()))
+        .collect::<Vec<_>>();
+
+    for (key, value) in define_entries {
+        for prefix_len in 1..key.len() {
+            let (prefix, suffix) = key.split_at(prefix_len);
+            if let Some((_, parent)) = defines
+                .iter_mut()
+                .find(|(candidate, _)| candidate.as_slice() == prefix)
+            {
```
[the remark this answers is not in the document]
fireairforce: done

crates/pack-core/src/shared/contexts.rs
```diff
@@ -14,23 +14,77 @@ use turbopack_core::{
...
+                if let Some((_, existing)) = entries.iter_mut().find(|(key, _)| key == segment) {
+                    *existing = value;
+                } else {
+                    entries.push((segment.clone(), value));
+                }
+                return true;
+            }
+
+            let child =
+                if let Some((_, existing)) = entries.iter_mut().find(|(key, _)| key == segment) {
```
xusd320, line 50: 这不是 hashmap 吗，为啥用 O(n) 的方法？

xusd320 (changes requested):

crates/pack-core/src/shared/contexts.rs
fireairforce: https://github.com/utooland/next.js/blob/utoo/turbopack/crates/turbopack-core/src/compile_time_info.rs#L126 这个是 vec

xusd320 (approved):

PR #2790, opened 2026-04-15
feat(pack): support `__webpack_public_path__`(webpack compitable)

fireairforce:
## Summary

align: https://webpack.js.org/guides/public-path/#on-the-fly

mako also support: https://github.com/utooland/utoo/blob/master/crates/mako/src/visitors/public_path_assignment.rs#L19

### DIFF
--- a/crates/pack-core/src/shared/transforms/webpack_public_path.rs
+++ b/crates/pack-core/src/shared/transforms/webpack_public_path.rs
@@ -1,0 +1,95 @@
+use anyhow::Result;
+use async_trait::async_trait;
+use swc_core::{
+    atoms::atom,
+    common::SyntaxContext,
+    ecma::{
+        ast::*,
+        visit::{VisitMut, VisitMutWith},
+    },
+};
+use turbopack::module_options::ModuleRule;
+use turbopack_ecmascript::{CustomTransformer, TransformContext};
+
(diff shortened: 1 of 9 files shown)


crates/pack-core/src/shared/transforms/webpack_public_path.rs
```diff
@@ -0,0 +1,95 @@
...
+    fn global_public_path_expr(&self, span: swc_core::common::Span) -> Expr {
+        Expr::Member(self.global_public_path_member(span))
+    }
+}
+
+impl VisitMut for WebpackPublicPathVisitor {
+    fn visit_mut_assign_expr(&mut self, assign_expr: &mut AssignExpr) {
+        match &assign_expr.left {
+            AssignTarget::Simple(SimpleAssignTarget::Ident(binding_ident))
+                if self.is_webpack_public_path(&binding_ident.id) =>
```
xusd320, line 71: Is swc Atom Eq compare to Atom faster than Atom  Eq compare to str  ? @copilot

xusd320 (approved):

crates/pack-core/src/shared/transforms/webpack_public_path.rs
xusd320: https://gemini.google.com/share/fd6ddc08c8ed

PR #2799, opened 2026-04-17
feat(pm): support file: protocol dependencies

elrrrrrrr:
## Summary

Adds `file:` protocol support to the package manager, mirroring the HTTP tarball work from #2784. Both flavours handled:

- **Local tarball** — `"foo": "file:./foo-1.2.3.tgz"` → extract
- **Local directory** — `"foo": "file:./local-pkg"` → copy tree (excludes `node_modules` / `.git`; 512 MiB cap)

Cache slot: `<cache>/<name>/_file_<sha256(abs_path)[:16]>/package/` — same shape as `_http_<url_hash>`, keyed on absolute source path so monorepo name-collisions stay distinct.

Transitive file: deps resolve against the parent's on-disk origin (recovered from `dist.tarball = file:<abs>` stamped by `finalize_non_registry_manifest`). File: declared inside a registry-published package surfaces an explicit Unsupported error rather than silently resolving against cwd.

## Refactor

- Extracted `crates/ruborist/src/resolver/tar.rs` — `TarEntry`, `scan_tarball`, `write_entries`, `gzip_decompress`, `MAX_UNCOMPRESSED_BYTES`. `http.rs` and `file.rs` both import from it.
- `resolver/common.rs::cache_slot(prefix, bytes)` — single implementation backing `http_cache_slot` and `file_cache_slot`.
- `pm/util/downloader.rs::slot_cache_lookup(name, slot)` — shared install-phase probe.

## Bug caught during review

Directory walker previously used `DirEntry::file_type().is_file() || is_symlink()` and then `fs::read()`. `file_type()` returns `is_file()==false` for symlinks, so symlinked *directories* were being read as raw bytes into the cache. Now uses a single `fs::metadata()` call per entry (follows symlinks, covers type + mode on unix in one stat). Broken symlinks warn-and-skip instead of aborting. Added two regression tests.

## pm-arborist reclassification

Old `SKIP_FILE_PROTOCOL` (18 fixtures, one bucket) split by actual root cause:

| Bucket | Fixtures | Why |
|---|---|---|
| `SKIP_FILE_TARGET_MISSING` | 8 | target dirs/tarballs weren't ported from `npm/cli` |
| `SKIP_FILE_SEMANTIC` | 3 | transitive file: inside registry; `./a` vs github shorthand; missing manifest name |
| `SKIP_BUNDLED_FILE` | 4 | bundledDeps-copy-from-file: not yet implemented |

Five fixtures unskipped: `tarball-dependencies`, `link-dep-nested`, `old-lock-with-link`, `testing-asymmetrical-bin-no-lock`, `workspaces-with-files-spec`.

## Test plan

- [x] `cargo test -p utoo-ruborist --features http-tarball resolver::file` — 9/9 pass (includes 2 new symlink regression tests)
- [x] `cargo fmt --check` + `cargo clippy --all-targets -- -D warnings --no-deps` (ruborist + pm) — clean
- [x] Local cold + warm install of `e2e/pm/file-deps/` — both pass
- [ ] CI: `e2e/utoo-pm.sh` Case 8.4 / 8.5
- [ ] CI: `e2e/pm-arborist.sh` — confirm the 5 unskipped fixtures pass

## Known limitations (documented in skip buckets)

- Lockfile stores absolute paths in `resolved`. BFS re-resolves every install so it self-heals across machines but dirties the lockfile. npm's relative-path serialization is future polish.
- `link:` / `portal:` still Unsupported (not in scope).

## Related

Follow-up to #2784 (HTTP tarball deps).


elrrrrrrr:
### Update: `file:<dir>` installs as symlink (not copy), tgz reuses http machinery

Two fixes per review:

**1. Directory deps are now symlinks, matching npm**

The first pass was wrong — npm (5+) and pnpm both symlink `file:./local-dir` deps in place; only `file:./pkg.tgz` involves extraction. Fixed:

- `resolve_dir_blocking` reads `package.json` only (no cache touch); stamps `manifest.dist.tarball = "link:<abs>"` as a sentinel.
- `create_non_root_lock_package` detects the sentinel and emits `link: true` + `resolved: <relative>` — same lockfile shape as workspace links.
- The existing install-phase link code path takes over; no changes needed there.
- `file_base_dir` in the builder now also accepts `link:<abs>` (the parent for a `link:` is the dir itself) so transitive `file:` deps inside a symlinked dir resolve right.

**2. Tarball path now reuses http almost entirely**

Pulled `decompress → scan → finalize → atomic-commit` into `tar::commit_tarball_bytes(cache_dir, bytes, pinned_url, slot)`. Both resolvers are thin wrappers:

```rust
// http — download then commit
fn fetch_and_extract_blocking(cache_dir, url, bytes) -> Result<Manifest> {
    commit_tarball_bytes(cache_dir, bytes.as_ref(), url.into(), &http_cache_slot(url))
}

// file tarball — read disk then commit
fn resolve_tarball_blocking(cache_dir, abs_src) -> Result<Manifest> {
    let bytes = fs::read(abs_src)?;
    commit_tarball_bytes(cache_dir, &bytes, format!("file:{}", abs_src.display()), &file_cache_slot(abs_src))
}
```

### Verified

- E2E now asserts `node_modules/local-dir-pkg` is a **symlink** to `../local-dir` and `node_modules/local-tarball-pkg` is a **real directory** clone. Lockfile assertions: dir entry has `link: true`, tarball entry has `resolved: file:<abs>`.
- All 7 file resolver unit tests pass; cargo fmt + clippy clean.
- Cold + warm install both work; the ~260 lines of directory-walking code (`walk_into` / `scan_directory` / `EXCLUDED_DIR_NAMES` / etc.) are gone since directories are no longer copied.

Commit: c6096cbe

elrrrrrrr:
### Update: lockfile `resolved` is now root-relative (npm-compatible)

Per review: confirmed npm's behavior in `~/code/cli/workspaces/arborist/lib/shrinkwrap.js` — `resolved` is always relative to the project root, never absolute. My earlier output had absolute paths for tarball deps.

Examples from npm fixtures (`workspaces/arborist/test/fixtures/external-link-dep/package-lock.json`):
- Link: `"resolved": "../cli-750"`, `"resolved": "node_modules/abbrev"`
- Tarball: `"version": "file:child-1.2.3.tgz"`

**Now matches**:

```json
"node_modules/local-dir-pkg":     {"link": true, "resolved": "local-dir"}
"node_modules/local-tarball-pkg": {"resolved": "file:local-tarball.tgz"}
```

**Implementation**:
- `get_relative_path` switched from `Path::strip_prefix` → `pathdiff::diff_paths` so out-of-root targets (e.g. `file:../sibling-dir`) serialize correctly (otherwise the strip would silently fall back to absolute).
- `dist.tarball` for tarball file: deps is rewritten from `file:<abs>` → `file:<rel>` at lockfile-serialization time via a small `rewrite_resolved` helper.
- The in-memory pipeline keeps using `file:<abs>` (BFS-time URLs); only the lockfile output is relative.
- Install side: plumbed project-root `cwd` through `clone_package_once → resolve_cache_path → file_cache_lookup`. When the lookup gets a `file:<relative>` URL from the lockfile, it re-absolutizes via `cwd.join()` before hashing the slot, so it agrees with what BFS wrote.

**Verified**:
- Cold + warm install both succeed with relative-path lockfile.
- E2E asserts the exact strings: `"resolved": "local-dir"` and `"resolved": "file:local-tarball.tgz"`.
- 7 file resolver unit tests pass; clippy clean.

Commit: d4dac8ae

fireairforce (approved):

fireairforce:
终于支持了

PR #2808, opened 2026-04-18
feat(pack): harden server bundling and simplify configuration

xusd320:
This PR hardens the RSC-like server bundling process and simplifies the configuration of server functions.

Key changes:
- Support for optional server entries.
- Unified server chunking: server entry and server functions are now bundled into a single `index.js` chunk.
- Configuration simplification: flattened `clientReference` and `serverReference` directly into the `server` configuration block.
- Improved path resolution: users can now use project-relative paths (e.g., `./input/transport.ts`) for client and server references, which are handled via internal bundler aliases.
- Refactored AST transformer to use fixed internal specifiers for server references.
- Prepares the framework to properly support the `examples/with-server-references` workspace.
- Updated snapshot tests to reflect these changes.
- Fixed clippy warnings in `pack-api/src/app.rs`.


xusd320:
@gemini-code-assist All feedback points addressed:
1. `app.rs:493` - Expanded conditional logic so the server bundle executes if `client_reference` OR `entry` is defined.
2. `app.rs:605` - Removed the early returns on empty `unique_server_modules`, ensuring resolution of the configured server entry even if no actual server functions exist in the client graph.
3. `proxy.rs:292` - Discarded the manual `default` block list during server reference AST registration, enabling consistent support for default exports.
4. `app.rs:682` - The server chunk resolution now strictly obeys the provided `server.output.filename` schema, resolving gracefully to `index.js` as a fallback.

fireairforce (approved):

PR #2810, opened 2026-04-19
feat(pack): keep dev watch active before page open

fireairforce:
## Summary

Keep devServer watch active even if the page not open by the user.


xusd320:
完全关掉 tracing ? 那 dev 的启动过程这么感知呢

xusd320 (approved):

PR #2816, opened 2026-04-20
perf(pm): DirFd-based I/O for cloner + extractor (-20% cold install)

elrrrrrrr:
## The problem

During a cold install, `utoo` does two million-scale file I/O passes over `node_modules`:

1. **Tarball extraction** — one `File::create(dest.join(rel_path))` per file
2. **Cache → node_modules hardlinking** — one `fs::hard_link(abs_src, abs_dst)` per file

Both hand the kernel full **absolute paths**, forcing a fresh dentry walk from `/` on every call. For a typical source/dest pair like `~/.cache/nm/lodash/4.17.21/package/src/foo.js → <proj>/node_modules/lodash/src/foo.js`:

```
oldpath: / home / user / .cache / nm / lodash / 4.17.21 / package / src / foo.js   ← 9 segments
newpath: / proj / node_modules / lodash / src / foo.js                              ← 6 segments
```

That's ~15 dentry lookups per `link(2)`. An ant-design-scale project has ~100k files × two syscall chains (extract + link), which is **millions of repeated dentry walks** — and most of the prefix is constant (cache root + node_modules root never change within a package). Beyond wasted kernel CPU, that repetition creates contention on dcache seqlocks and `rename_lock` under the parallel install.

## The fix

Introduce a `DirFd` abstraction (`util/at.rs`, thin rustix wrapper) that holds an opened directory fd. I/O underneath it goes through **relative paths**:

```rust
// Open parent dirs once
let src_dir = DirFd::open("/home/user/.cache/nm/lodash/4.17.21/package")?;
let dst_dir = DirFd::open("/proj/node_modules/lodash")?;

// Every subsequent link resolves one component
dst_dir.link_from(&src_dir, c"foo.js")?;  // linkat(src_fd, "foo.js", dst_fd, "foo.js", 0)
dst_dir.link_from(&src_dir, c"bar.js")?;
dst_dir.link_from(&src_dir, c"baz.js")?;
```

Kernel semantics: **`linkat(dirfd, name, ...)` resolves `name` relative to the dentry `dirfd` points at, not `/`**. Passing a single-component name (no slash) does exactly one dentry lookup on each side. The cost of the one-time `openat` is amortized over every file in the directory — and npm packages are exactly the "one directory holds tens to hundreds of files" shape that this pattern fits.

The same principle applies to tarball extraction: open every intermediate directory once, then create leaf files via `openat(parent_fd, "leaf", O_CREAT | ...)`.

#### `util/at.rs` — new `DirFd` (Linux-only)

```rust
pub struct DirFd(OwnedFd);

impl DirFd {
    pub fn open(path: &Path) -> io::Result<Self>;
    pub fn open_child(&self, name: &CStr) -> io::Result<Self>;
    pub fn mkdir(&self, name: &CStr, mode: u32) -> io::Result<()>;
    pub fn link_from(&self, src: &Self, name: &CStr) -> io::Result<()>;
    pub fn create_file(&self, name: &CStr, mode: u32) -> io::Result<OwnedFd>;
    pub fn read_entries(&self) -> io::Result<Dir>;
    pub fn stat(&self, name: &CStr) -> io::Result<Stat>;
}
```

All seven methods have real call sites in cloner / extractor. rustix handles the u32/u16 cross-platform `RawMode`, `io::Errno` → `io::Error` conversion, and the `fdopendir` dup dance — no `unsafe` in our code.

#### `util/cloner.rs` — Linux hardlink path rewritten on `DirFd`

The old libc implementation (collect → mkdir phase → link phase) collapses into a single recursive walk:

```rust
fn clone_dir_at(src_dir: &DirFd, dst_dir: &DirFd, ..) -> io::Result<()> {
    for entry in src_dir.read_entries()? {
        let name = entry.file_name();
        match entry.file_type() {
            Directory => {
                dst_dir.mkdir(name, 0o755)?;
                clone_dir_at(&src_dir.open_child(name)?, &dst_dir.open_child(name)?, ..)?;
            }
            _ => {
                dst_dir.link_from(src_dir, name)?;  // hot path: one syscall, single-component both sides
            }
        }
    }
}
```

All existing behavior is preserved:
- **EXDEV latch** — cross-device hardlink fails as a property of the src/dst pair, so the first failure latches `force_copy` and the rest of the clone falls back
- **Per-file copy fallback** — EMLINK / EPERM are per-inode or per-file and don't latch
- **`_hasInstallScript` force-copy** — packages with install scripts never hardlink, so postinstall mutations can't leak back into the shared cache
- **`DT_UNKNOWN` → `fstatat`** — older XFS / ReiserFS don't populate `d_type`

#### `util/extractor.rs` — tarball extraction via `DirFd`

Previous flow: decompress → collect entries (with absolute paths) → sort dirs shallow-first + `create_dir` each → rayon `par_iter` writing each file with `File::create(abs)`.

New flow on Linux:

1. `DirFd::open(dest)` once at entry
2. Walk `rel_dirs` shallow-first, `parent.mkdir(leaf, 0o755)` + `parent.open_child(leaf)`, store the resulting DirFds in a `HashMap<PathBuf, Arc<DirFd>>`
3. rayon parallel write: look up the parent DirFd, `parent.create_file(leaf, mode)` → `std::fs::File::from(fd)` → `write_all`

Two extra wins:
- File mode is set via `openat(O_CREAT, mode)` at creation time, eliminating the **separate `chmod` pass** the old code ran for non-`0o644` files
- `_resolved` marker creation goes through the same DirFd, keeping the path consistent

#### What this PR does NOT change

- **macOS**: `clonefile` is already a single-syscall whole-tree copy — it's optimal, untouched
- **Windows**: `CreateHardLink` has no `At` variant, so `fs::hard_link` stays as-is
- **`validate_directory`**: only runs on the warm / reinstall path (dst already exists); the bench showed no meaningful warm-install delta from a DirFd version, so the original tokio async impl is kept
- **`.bin` symlinks** via `symlinkat`: 1–2 entries per package, too small to matter — future work

#### On fd pressure

Per-clone peak fd usage is `3 × (recursion depth + 1)` — `src_fd`, `dst_fd`, plus the dup'd fd that backs `DirIter`'s `fdopendir`. Typical npm package trees are 3–5 deep, so a single clone holds 15–30 fds. With tokio `spawn_blocking`'s 512-thread ceiling, worst case is ~15k fds. `util/sysconf.rs::raise_fd_limit` already bumps `RLIMIT_NOFILE` from soft to hard at startup (Linux runners default to 65535+), so there is headroom.

## Benchmarks (Ubuntu runner · 5 cold / 10 warm runs)

| Scenario | utoo-npm | utoo-next | **utoo (this PR)** | vs next |
|---|---|---|---|---|
| cold ant-design-x / npmjs | 14.10 ±2.22 | 14.73 ±3.80 | **12.28 ±0.39** | **-17%** |
| cold ant-design-x / npmmirror | 35.40 ±29.60 | 28.17 ±10.74 | **21.63 ±0.26** | **-23%** |
| cold ant-design / npmjs | 15.64 ±3.18 | 15.94 ±5.14 | **11.28 ±0.95** | **-29%** |
| cold ant-design / npmmirror | 20.91 ±6.26 | 23.18 ±7.18 | 20.40 ±3.98 | -12% |

- **Cold install ~20% faster on average**
- **Error bars shrink significantly** (e.g. ant-design-x/npmjs σ drops from ±3.80 to **±0.39**) — fewer path-walk-induced scheduling spikes
- Resources table: `IO reads` drop to ~0 (from hundreds/thousands), voluntary context switches down ~15%, consistent with the reduced dentry lookup count
- Warm install doesn't touch the extractor; observed delta is within noise

## Scope

| File | Net lines added |
|---|---|
| `at.rs` | +92 |
| `cloner.rs` | +146 |
| `extractor.rs` | +142 |
| Cargo / mod / typos | ~9 |
| **Total** | **+351 / -38** |


elrrrrrrr:
## Closing without merging — bench data doesn't justify the complexity

Final verdict after multiple bench rounds on Linux + macOS:

**Wall-clock cold install:** utoo (this PR) vs utoo-next fluctuated between -11% and +3% across runs, which turned out to be within GitHub Actions runner variance (utoo-npm — an identical published binary — varied 11% across runs for the same workload). The earlier "stable -20%" numbers were lucky samples at 3-run resolution. With 5–10 runs per cell and multiple repeats, no consistent wall-clock win remains.

**Kernel-level win is real but invisible to users:** voluntary context switches drop ~10% on Linux (200k → 180k for ant-design-x / npmjs cold). The `openat` + `linkat` path demonstrably reduces dentry-walk work per file. But this slice is ~10–15% of total cold-install time; the other 85% (network download, gzip decompress, tar parse, buffer copies, tokio↔rayon boundaries) dominates.

**macOS is worse:** APFS has no Linux-style dentry-walk overhead, so the `DirFd`/`openat` path is pure added machinery. Bench shows +13% to +50% regressions with 3–4× σ increase.

**bun's 2–3× lead is architectural, not syscall-level:** bun does ~21k voluntary ctx switches vs utoo's ~200k on the same workload. Closing that gap requires bigger structural changes — stream decompress, single-pass write (skip cache→node_modules double write), eliminate tokio↔rayon oneshot handoff. None of which fit in this PR.

**What stays valuable as follow-up:**
- The `data-uri-to-buffer@6.0.2` / `degenerator@5.0.1` dup-tar-entry race is a real bug that would bite any future `openat`-based extractor — worth documenting.
- Ctx-switch optimization is the direction, not syscall-level tuning. Separate PRs per direction (stream decompress, merge cache+install write, tokio/rayon boundary reduction).

Closing. Thanks for the ride.

Issue #2811, opened 2026-04-19
[Utoo install] Nested workspace packages can't resolve declared deps — hoisted to anchor only


elrrrrrrr:
### Describe the bug

When `ut install --from pnpm` anchors at a non-root workspace (a `package.json` / `pnpm-workspace.yaml` that is not the repo root), deps get hoisted into the anchor's `node_modules/`. Nested workspace packages referenced by that anchor (e.g. `../../packages/foo`) **cannot resolve those hoisted deps** — Node's module resolution walks up from the script location, and the anchor's `node_modules/` is sideways, never traversed.

Same root cause manifests twice:

1. Duplicate-package postinstall failure (`Cannot find module '@electron/get'`)
2. Any declared dep of a nested workspace package fails to resolve (25+ TS2307 in our repo)

### Reproduction

```shell
git clone https://github.com/elrrrrrrr/lobehub -b ut-ci-test
cd lobehub/apps/desktop
ut install --from pnpm     # anchors at apps/desktop (7 workspaces)
```

Observed:

```
Error: Cannot find module '@electron/get'
  at packages/electron-client-ipc/node_modules/electron/install.js:3
```

Layout:

- `apps/desktop/node_modules/electron@41.1.0` + `apps/desktop/node_modules/@electron/get` ✓
- `packages/electron-client-ipc/node_modules/electron@39.8.8` (nested due to version conflict)
- `packages/electron-client-ipc/node_modules/electron/node_modules/` contains only `@types`, **no `@electron/get`**

Node resolves upward from `install.js` → never reaches `apps/desktop/node_modules/` (sideways).

### Expected

For each nested workspace package, its declared deps must be reachable via Node's upward resolution. Three valid layouts:

- **A.** Nested: `packages/<x>/node_modules/<dep>/node_modules/<subdep>`
- **B.** Co-located: `packages/<x>/node_modules/<dep>`
- **C.** Symlink to hoisted copy (pnpm's `.pnpm/` approach)

utoo currently appears to dedupe to the anchor without checking reachability from nested packages.

### Environment

- utoo `1.0.27`
- ubuntu-24.04 (GitHub Actions) and macOS — both reproduce
- Node 24.x

PR #2819, opened 2026-04-21
fix(pm): avoid reusing anchor-only hoists for workspaces

killagu:
## Summary
- resolve dependency reuse from the requester path instead of blindly starting at the graph parent
- only reuse hoisted packages from node_modules directories that are actually reachable by Node upward resolution
- add regression tests for nested-preferred reuse and sibling workspace hoist isolation

## Validation
- cargo +nightly-2026-02-05 fmt --check
- cargo +nightly-2026-02-05 test -p utoo-ruborist test_find_compatible_node_ -- --nocapture

## Reproduction
- reproduced on elrrrrrrr/lobehub branch ut-ci-test after bypassing an unrelated HTTP tarball blocker
- confirmed apps/desktop/node_modules/xlsx existed while packages/file-loaders could not resolve xlsx from its own directory

### DIFF
--- a/crates/ruborist/src/model/graph.rs
+++ b/crates/ruborist/src/model/graph.rs
@@ -438,12 +438,15 @@ impl DependencyGraph {
         } else {
             version_spec.to_string()
         };
+        let requester_path = self
+            .get_node(from)
+            .map(|node| normalize_lexical_path(&node.path))
+            .unwrap_or_default();
 
-        // Get physical parent of from node, default to from if it's the root
-        let parent = self.get_physical_parent(from).unwrap_or(from);
-
-        // Recursively search up the parent chain
-        self.find_in_parent_chain(parent, name, &effective_spec, from)
+        // Search from the requester itself, then walk up through only the
+        // ancestors whose `node_modules/` directories are actually reachable
+        // via Node's upward resolution from the requester's on-disk path.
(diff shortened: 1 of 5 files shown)


killagu:
Addressed the inline review comments:
- carry a normalized requester path through find_in_parent_chain instead of repeatedly looking it up by node index
- normalize both requester/candidate paths lexically before the reachability check so dot-dot segments cannot produce false reachable results
- switch the sibling-workspace regression test to platform-agnostic relative paths

Revalidated with:
- cargo +nightly-2026-02-05 fmt --check
- cargo +nightly-2026-02-05 test -p utoo-ruborist test_find_compatible_node_ -- --nocapture

elrrrrrrr:
@killagu 需要补充一下原始 issue 里的 e2e case

elrrrrrrr:
https://github.com/utooland/utoo/issues/2811

原始 issue 在这，pr 里应该保持关联

elrrrrrrr:
Thanks @killagu — really appreciate the careful workspace-hoisting fix here. It's gone stale against current `next` (conflicting after the resolver/hoist changes that have since landed), so I'm closing it in a PM backlog cleanup rather than letting it linger.

If the anchor-only-hoist reuse is still reproducible on latest `next`, please do reopen or push a rebase — happy to help get it in. Thanks again for digging into this one.

PR #2818, opened 2026-04-21
perf(pm): sequential writes within tarball extraction

elrrrrrrr:
## Summary

Started as "intra-package sequential tarball writes" (single-line perf tweak) and evolved into a multi-week, data-driven preload-resolver overhaul plus a complete bench infrastructure. **This PR is now too large to review safely** and is being split — see "PR split plan" at the bottom.

This description captures the full exploration journey so subsequent split-PRs can reference it as context.

## End-to-end results (ant-design / npmjs.org / GitHub Actions ubuntu)

| Phase | bun | utoo (start) | utoo (now) | utoo Δ | vs bun |
|---|---:|---:|---:|---:|---:|
| **p0_full_cold** | 9.07-9.32s | ~14s | **8.89-9.06s** | **−5s** | **utoo wins / matches** |
| **p1_resolve** | 1.98-2.34s | 5.78s | **2.65s** | **−54%** | gap +0.31-0.67s |
| **p3_cold_install** | 7.71-7.96s | ~9s | **5.74s** | **−36%** | **utoo +2s ahead** |
| **p4_warm_link** | 3.42-3.51s | ~3s | **1.98s** | **−34%** | **utoo +1.5s ahead** |

utoo now matches or beats bun on 3 of 4 phases. The remaining p1 gap (+0.4s) is fully accounted for and at the architectural floor.

## Journey timeline — p1_resolve preload wall

```
5.78s baseline
  → 4.47s  eager Versions parse → kept the parse, refactored architecture
  → 4.09s  inline extract_transitive_deps + direct deserialize
  → 3.82s  lock-free SegQueue pending queue (was Mutex<VecDeque>)
  → 3.30s  rustls + aws-lc-rs (replaces ring; CCS→AppData 154ms→17ms)
  → 3.10s  tracing file_filter info+ (drop debug! eval cost)
  → 2.96s  Arc<FullManifest> in cache (no more deep clone)
  → 2.65s  drop indicatif Mutex from per-completion path
  → 2.15s  worker-pool replaces FuturesUnordered ⭐ biggest single win
  → 2.65s  current (Cloudflare variance: best 2.04s, typical 2.65s incl BFS+lockfile)
```

### 1. Worker-pool preload (commit `ed7b551e`) — the core win

[...]

PR #2825, opened 2026-04-25
perf(pm): chunked parallel writes within tarball extraction + Tar Slip guard

elrrrrrrr:
## Summary
Second of 4 split PRs from #2818. Single-file change in `crates/pm/src/util/extractor.rs` covering two related improvements to the install-path tarball extraction:

1. **Chunked parallel writes** — replace per-file `par_iter` with `par_chunks(32)`, cutting rayon task count 32× while preserving IO-overlap parallelism across cores
2. **Tar Slip guard** — reject entries whose path is absolute or contains `..` components

## Performance impact
CI p3_cold_install: utoo **5.74s** vs bun **7.71s** (utoo ~2s ahead).

A/B verified essential: removing intra-package `par_chunks` (sequential `for entry in &entries` inside each rayon task) regressed p3 **+3.67s** and exploded σ from 0.04 → 2.85 (IO can't interleave across cores).

## Why these go together
Single function change — both edits touch the same `extract_tarball_sync` body. Splitting further would force a no-op intermediate state that doesn't compile cleanly.

## Test plan
- [x] `cargo fmt` + `cargo clippy --all-targets -- -D warnings --no-deps` clean
- [x] `cargo test -p utoo-pm` 249 passed
- [ ] CI ant-design install on `pm-bench-phases` workflow (will run once #2824 lands and bench-phases label can be applied here)

## Context
- Full exploration journey + failed-experiments catalog: #2818
- Bench infrastructure (auto-comments p0/p1/p3/p4 deltas on labeled PRs): #2824 ⬅ recommend reviewing first

### DIFF
--- a/.github/actions/setup-utoo-next-baseline/action.yml
+++ b/.github/actions/setup-utoo-next-baseline/action.yml
@@ -1,0 +1,23 @@
+name: setup-utoo-next-baseline
+description: Download the utoo binary built from origin/next HEAD (by build-next-* jobs) into a temp directory and export $UTOO_NEXT_BIN. Used by bench-phases to add a `utoo-next` baseline column — the merged-but-not-published code — so PRs can show their delta vs the trunk baseline (not just vs the older npm-published version).
+
+inputs:
+  artifact:
+    description: Artifact name uploaded by the matching build-next-* job (e.g. utoo-next-linux-x64)
+    required: true
+
+runs:
+  using: composite
+  steps:
+    - name: Download utoo-next binary
+      uses: actions/download-artifact@v4
+      with:
+        name: ${{ inputs.artifact }}
+        path: /tmp/utoo-next-pkg
+    - name: Set UTOO_NEXT_BIN
+      shell: bash
+      run: |
+        UTOO_NEXT_BIN="/tmp/utoo-next-pkg/utoo"
+        chmod +x "$UTOO_NEXT_BIN"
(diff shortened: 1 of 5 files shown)


elrrrrrrr:
## Closing — clean A/B doesn't show the expected gain

After the bench infra got a `utoo-next` baseline column (next-branch HEAD, clean A/B), the chunked-write change measured against the trunk baseline shows no real improvement and a clear regression on npmmirror:

**linux p3_cold_install** (the phase this change targets):

| Registry | utoo-next (trunk baseline) | utoo (this PR) | Delta |
|---|---|---|---|
| npmjs | 7.72s ±1.64s | 7.55s ±2.28s | **-0.17s** (within σ — no improvement) |
| npmmirror | 21.67s ±0.47s | 24.95s ±1.25s | **+3.28s** (clear regression, well outside σ) |

The earlier "vs utoo-npm faster by 1.4s" reading on this PR was misleading — utoo-npm @ 1.0.27 was published 10 days ago, so the gap conflated this PR with 6 other PRs merged onto next since then. The previous A/B from #2818 ("removing par_chunks regressed +3.67s") was run on a parent perf branch carrying additional commits — re-tested standalone here, the chunked-write change doesn't carry its weight.

Closing rather than tweaking — `entries.par_iter().for_each(write)` is fine as-is. The Tar Slip guard mentioned in the original commit is already on next via #2665.

PR #2839, opened 2026-04-27
fix(pm): run install lifecycle on workspace packages + unify lifecycle primitive

### LINKED ISSUE #2833: ut install / ut rebuild does not run `prepare` (or `postinstall`) for workspace packages — diverges from npm/pnpm/tnpm
## Summary

`ut install` (and `ut rebuild`) does **not** invoke the `prepare` lifecycle script of workspace packages after linking them, so a workspace package that needs a build step before it can be consumed by another workspace package leaves consumers broken until the user manually builds it.

This diverges from npm 7+, pnpm, yarn, and tnpm, all of which run `prepare` for every workspace package once installation/linking completes (npm RFC: [scripts spec](https://docs.npmjs.com/cli/v10/using-npm/scripts#prepare-and-prepublish), npm 7 workspaces docs).

## Version

- utoo: `1.0.27`
- platform: macOS (darwin 25.1.0)

## Repro

In a monorepo where workspace package `B` depends on workspace package `A`, and `A` ships its build output (no source-level entry):

```jsonc
// packages/A/package.json
{
  "name": "A",
  "main": "lib/index.js",
  "scripts": {
    "build": "tsc -p .",
    "prepare": "npm run build"
  }
}

// packages/B/package.json
{
  "name": "B",
  "dependencies": { "A": "workspace:*" },
  "scripts": { "compile": "webpack" } // imports from A's lib/
}
```

Steps:

```bash
rm -rf packages/A/lib node_modules
ut install
cd packages/B && ut compile      # ← fails: Module not found A/lib/...
```

Expected: \`packages/A/lib\` exists after \`ut install\`, because A's \`prepare\` should have been invoked.
Actual: \`packages/A/lib\` does not exist; consumer build fails.

For comparison, \`tnpm update\` (delete node_modules + reinstall) on the same tree builds \`A/lib\` correctly via \`prepare\`.

I also tried:
- \`ut rebuild\` — does not run lifecycle scripts on workspace source packages either.
- Renaming the script to \`postinstall\` — also not invoked.

## Real-world reproduction

Repo: \`ant-design/ant-design-icons\` master branch.

\`\`\`bash
git clone https://github.com/ant-design/ant-design-icons
cd ant-design-icons
ut install
cd packages/icons-react
tnpm run prepublishOnly   # works
ut run prepublishOnly     # also works because pregenerate hook is now wired,
                          # BUT without the pregenerate workaround it fails:
                          # Module not found: '@ant-design/icons-svg/lib/asn/AccountBookFilled'
\`\`\`

The root cause is that \`@ant-design/icons-svg\` is consumed by \`@ant-design/icons\` via its built \`lib/asn/*\` files; those are produced only by its build step. tnpm's \`prepare\` semantics handle this transparently, but \`ut install\` leaves \`lib/\` empty.

## Suggested fix

After completing workspace linking in \`ut install\`, walk workspace packages in topological order and execute \`prepare\` (and \`preinstall\`/\`install\`/\`postinstall\` for non-workspace deps as today). Same for \`ut rebuild\` — npm's \`npm rebuild\` runs install scripts; \`prepare\` should be triggered on workspace packages there too.

Workaround for monorepo authors today: declare a \`pre<script>\` hook on the consumer that explicitly builds the upstream workspace package, e.g.:

\`\`\`json
"pregenerate": "cd ../icons-svg && ut compile"
\`\`\`

But this is awkward and breaks the standard npm contract that downstream tools (including tnpm/pnpm/yarn) honor.

## Related

- #2811 — also about workspace edge cases in \`ut install\`.

elrrrrrrr:
## Summary

`ut install` / `ut rebuild` skipped lifecycle install hooks (`prepare`, `postinstall`, etc.) on workspace source packages, so a workspace consumer that imported another workspace's build output broke until the upstream package was built manually. This diverged from npm 7+, pnpm, yarn, and tnpm.

After `node_modules` is linked, walk workspace packages in topological order and run npm's install lifecycle (`install` / `prepublish` / `prepare` event chains) on each, **before** the root's own hooks — the root may import build artifacts produced by a workspace's `prepare`.

While here, fold two related cleanups in (small enough to ship together):
1. Normalize anonymous workspace names (`@npmcli/name-from-folder` fallback) at discovery time so downstream consumers see a non-empty `pkg.name` without parallel `_with_name` constructors.
2. Collapse the duplicate `pre<event>` / `<event>` / `post<event>` chains into a single `ScriptService::run_lifecycle` primitive used by both `run` and the install path.

Closes #2833.

## Behaviour changes (the fix)

- `PackageService::process_workspace_install_hooks` resolves the workspace topology and runs each workspace's npm install lifecycle in order, gated on `ScriptPolicy::Run` so `--ignore-scripts` semantics are preserved.
- `RebuildService::rebuild` invokes the workspace lifecycle walk before `process_project_hooks`.
- Anonymous workspaces (no `name` field — e.g. arborist's `workspaces-need-update` fixture) get the npm folder-derived fallback inside `find_workspaces`, so all downstream code sees a non-empty `pkg.name`.

## Refactors folded in

**Anonymous workspace name normalization** (`crates/ruborist/src/resolver/workspace.rs`)

The `name_from_folder` fallback now writes back into the in-memory `WorkspacePackage.package_json.name`. Removes the parallel `_with_name` constructors that an earlier draft sprinkled across `tree_builder`, `model::package`, and `ruborist::model::graph` — those files are untouched now.

**`ScriptService::run_lifecycle` primitive** (`crates/pm/src/service/script.rs`)

One function expresses "run `pre<event>` / `<event>` / `post<event>` for a single package" parameterised by:
- `LifecycleSink::Stream { workspace_label }` — terminal streaming (used by `run_script` and the install path).
- `LifecycleSink::Capture { workspace_label, header, body }` — caller-owned buffers; on partial failure the caller still has the bytes captured up to the failing step (used by `run_script_captured`/`run_in_layers`).
- `main_required: bool` — whether a missing main script is an error (`true` for `npm run <name>`, `false` for npm install events where any of the three may exist alone).

Callers shrink to thin wrappers:

| Caller | Before | After |
|---|---|---|
| `run_script` | 50 lines, hand-rolled pre/main/post | `run_lifecycle(Stream, main_required: true)` |
| `run_script_captured` | 100 lines, hand-rolled | `run_lifecycle(Capture, main_required: false)` + friendly missing-script CLI hint |
| `process_workspace_install_hooks` | 7-hook for-loop calling `execute_script` | 3 npm events × `run_lifecycle` via shared `run_install_lifecycle` helper |
| `process_project_hooks` | 7-hook for-loop | same helper |

Drops the `LifecycleHook::PROJECT_INSTALL_HOOKS` constant (no remaining users). Dependency-side `execute_script_queue` (per-hook slicing + `join_all`) and the publish/pack lifecycles keep using `execute_script` since they address single hooks by enum, not npm events.

## Changes

- `crates/ruborist/src/resolver/workspace.rs` — write `name_from_folder` fallback back into `WorkspacePackage.package_json.name`; add `test_name_from_folder` covering plain / @scope / root paths.
- `crates/pm/src/service/script.rs` — add `LifecycleSink` + `run_lifecycle`; rewrite `run_script` and `run_script_captured` as thin wrappers.
- `crates/pm/src/service/package.rs` — `process_workspace_install_hooks` resolves topology + loads workspaces with normalized names; `process_project_hooks` and `process_workspace_install_hooks` route through new `run_install_lifecycle` helper using `NPM_INSTALL_EVENTS = ["install", "prepublish", "prepare"]`.
- `crates/pm/src/service/rebuild.rs` — invoke workspace lifecycle walk before project hooks, gated on `ScriptPolicy::Run`.
- `crates/pm/src/model/package.rs` — drop `LifecycleHook::PROJECT_INSTALL_HOOKS`.

## Test plan

- [x] New unit tests in `crates/pm/src/service/package.rs`:
  - `test_process_workspace_install_hooks_topological_order` — B depends on A; A's `prepare` produces `lib/marker`; B's `prepare` asserts the marker exists.
  - `test_process_workspace_install_hooks_no_workspaces` — single-package project is a no-op success.
  - `test_process_workspace_install_hooks_propagates_failure` — a failing workspace `prepare` surfaces as an install error.
  - `test_process_workspace_install_hooks_anonymous_workspaces` — both anonymous siblings run their hooks via the folder-name fallback.
- [x] `crates/ruborist/src/resolver/workspace.rs` — `test_name_from_folder` for plain / `@scope/foo` / root paths.
- [x] e2e cases (`e2e/utoo-pm.sh`):
  - workspace-prepare: lib-a prepare → lib-b prepare → app prepare topological order, plus postinstall marker
  - workspace-prepare with `--ignore-scripts`: hooks skipped, bin link still runs
  - workspace-anonymous: name-from-folder fallback runs both anonymous siblings
- [x] Manual verification against `target/release/utoo`:
  - `ut run build --workspaces` (layered output unchanged)
  - `ut run build --workspace lib-b` (single-workspace layered path)
  - `ut run test --workspaces --if-present` (silent skip)
  - `ut run nonexistent --workspaces` (Fail-mode missing-script error preserved)
- [x] `cargo test -p utoo-pm -- --test-threads=1` — 249 passed (parallel runs hit a pre-existing cwd race in unrelated tests, not caused by this PR)
- [x] `cargo fmt --check` clean
- [x] `cargo clippy --all-targets --no-deps -- -D warnings` clean


killagu (approved):
LGTM

PR #2844, opened 2026-04-27
fix(pm): drain pipeline worker tasks

killagu:
## Summary
- Track pipeline download/clone subtasks with JoinSet and drain them before worker completion.
- Bound pipeline task fanout by the existing manifest concurrency setting.
- Avoid waiting forever for local/workspace parent clone keys that are never registered.

## Verification
- cargo fmt --check
- cargo test -p utoo-pm

Note: full cargo clippy --all-targets -- -D warnings --no-deps is blocked in this environment because openssl-sys requires pkg-config/OpenSSL for non-PM workspace targets.


elrrrrrrr:
Closing as stale: this draft is a one-off agent experiment from 2026-04-27 with no follow-up, and overlaps with sibling PRs exploring the same optimization. Reopen if revisited.

PR #2847, opened 2026-04-27
perf(pm): speed up package clone materialization

killagu:
## Summary
- Prefer the standard extracted tarball package/ directory without scanning cache directories for every clone.
- Avoid recursive create_dir_all for every collected package subdirectory; only the clone root uses create_dir_all and children use create_dir.
- Add coverage for the package/ fast path.

## Verification
- cargo fmt
- cargo test -p utoo-pm
- BENCH_RUNS=1 PM_LIST=utoo REGISTRY=https://registry.npmmirror.com bash bench/pm-bench-phases.sh (local utoo release-local binary first in PATH)

## Notes
- Full workspace clippy was attempted but the environment is missing pkg-config/OpenSSL for unrelated pack dependencies (openssl-sys build script).

### DIFF
--- a/crates/pm/src/util/cloner.rs
+++ b/crates/pm/src/util/cloner.rs
@@ -112,7 +112,6 @@ use libc::clonefile;
 
 #[cfg(not(target_os = "macos"))]
 mod hardlink_clone {
-    use std::collections::HashSet;
     use std::path::{Path, PathBuf};
     use std::{fs, io};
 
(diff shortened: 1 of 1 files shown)


elrrrrrrr:
Closing as stale: this draft is a one-off agent experiment from 2026-04-27 with no follow-up, and overlaps with sibling PRs exploring the same optimization. Reopen if revisited.

PR #2859, opened 2026-04-28
perf(pm): chunked parallel writes within tarball extraction

elrrrrrrr:
## Summary

Second cherry-pick from #2818 (perf/extract-sequential-writes), targeting **install path / p3_cold_install**. Restructures intra-package tar entry writes in `crates/pm/src/util/extractor.rs` to reduce rayon work-stealing overhead while keeping multi-core IO overlap for large packages.

Two commits, applied as a sequence (not squashed) to preserve the empirical "why":

1. **`perf(pm): sequential writes within tarball extraction`** — Replace `par_iter` with sequential `for` loop. Each tar entry is small (microseconds to write) so per-entry rayon tasks were dominated by futex park/unpark. Worse than running the writes inline.

2. **`perf(pm): chunked parallel writes within tarball extraction`** — Replace the sequential loop with `par_chunks(WRITE_CHUNK_SIZE)`, where each rayon task writes a contiguous run of 32 files sequentially. Restores multi-core overlap for large packages while cutting task count by ~32×.

Together the two are the local optimum after the journey timeline tested both extremes:
- per-entry par_iter (too many tasks, work-stealing dominates)
- pure sequential (single-core IO bottleneck on large packages)
- `par_chunks(32)` (current — verified essential by reverting; sequential write regressed p3 by +3.67s with σ exploding 0.04 → 2.85)

### Files

`crates/pm/src/util/extractor.rs` only (49 lines changed). No new dependencies, no API changes, no Cargo.toml touched.

### Risk

Low. Confined to one file in a hot path that's covered by `cargo test -p utoo-pm` and `e2e/utoo-pm.sh`. The chunk size constant `WRITE_CHUNK_SIZE = 32` is documented inline; tuning it requires a follow-up.

## Test plan
- [x] `cargo fmt`, `cargo clippy --all-targets -- -D warnings --no-deps -p utoo-pm` clean
- [x] `cargo build --profile release-local -p utoo-pm` clean
- [x] `tombi format --check`, `npx biome ci`, `typos` — all clean
- [ ] CI `pm-bench-phases` (label: benchmark) — verify p3_cold_install impact on Linux x86_64 / macOS

### DIFF
--- a/crates/pm/src/util/extractor.rs
+++ b/crates/pm/src/util/extractor.rs
@@ -65,17 +65,23 @@ async fn extract_tarball(gzip_bytes: Bytes, dest: &Path) -> Result<()> {
     rx.await.with_context(|| "Extract task panicked")?
 }
 
-/// Synchronous extraction: decompress + parse + sequential write, all on rayon.
+/// Synchronous extraction: decompress + parse + chunked parallel write, all on rayon.
 ///
-/// Writes are sequential within a package: tar entries are naturally streamed in order,
-/// and files are typically small (< 64KB). Intra-package parallelism via `par_iter`
-/// creates many short tasks that trigger rayon work-stealing (futex park/unpark) and
-/// dominate context switches on large dep graphs. Cross-package parallelism is preserved
-/// by the outer `rayon::spawn` in `extract_tarball`.
+/// Writes are batched into fixed-size chunks: each rayon task writes a contiguous
+/// run of `WRITE_CHUNK_SIZE` files sequentially. This keeps multi-core parallelism
+/// for IO-overlap while cutting the rayon task count (and its work-stealing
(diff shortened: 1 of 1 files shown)


elrrrrrrr:
Closing: local + Linux CI bench show no measurable improvement on ant-design / npmmirror. Δ median ±0.65s on macOS local (σ ≥2s), +0.05s on Linux CI npmjs (σ ≥1s) — well within noise. Journey timeline's claimed -3.67s appears to be a workload- or environment-specific local optimum, not portable across the bench-phases setup. The change may still help on ant-design-x scale or other platforms, but not enough signal to land it in isolation. Will re-evaluate as part of the larger #2818 split if/when worker-pool lands and amplifies extractor parallelism patterns.

PR #2867, opened 2026-04-29
perf(pm): switch utoo-pm to mimalloc global allocator

elrrrrrrr:
## Summary

Wire `mimalloc` as the global allocator for the `utoo`/`ut` binary. Pack (pack-cli, pack-napi, utoo-wasm) already uses mimalloc via `turbo_tasks_malloc::TurboMalloc` with the `custom_allocator` feature; PM was the lone holdout still on the system allocator.

The workspace already pins both `mimalloc` and `libmimalloc-sys` to the `utooland/mimalloc_rust` fork via `[patch.crates-io]`, so this reuses the existing dep graph — no new git source is added.

## Hypothesis

For PM workloads, allocation is concentrated in:

- **p1 (resolve)** — JSON manifest parsing (one document per package version, lots of small `String` / `Vec` allocations), dependency graph construction (`HashMap<String, ...>`, `Vec<NodeRef>`, `Arc::new`).
- **p0 (full cold install)** — everything in p1, plus tarball extraction (per-entry path/buffer allocations), filesystem layout planning, and rayon worker scratch.

Both phases are syscall- and CPU-bound rather than purely IO-bound, so a faster allocator (lower per-alloc overhead, better cache behavior, decommit pacing) typically shows up most clearly there.

## Change

- `crates/pm/Cargo.toml`: add `mimalloc = { version = "0.1", default-features = false }`.
- `crates/pm/src/main.rs`: register `MiMalloc` as `#[global_allocator]`.
- `Cargo.lock`: regenerated (the committed lockfile was already stale vs. source — the previous `perf(pm)` PR shows the same churn).

`cargo check` and `cargo clippy --all-targets -D warnings --no-deps` both clean on `crates/pm`.

## Test plan

- [ ] `bench-phases-linux` posts a sticky comment with p0/p1/p3/p4 numbers vs. bun (`benchmark` label triggers it on this draft).
- [ ] `pm-ci` (`(pm)` in title) green for unit tests + e2e.
- [ ] If p0/p1 improve and there's no regression elsewhere, mark ready for review; otherwise close.

### DIFF
--- a/.github/workflows/pm-e2e-bench.yml
+++ b/.github/workflows/pm-e2e-bench.yml
@@ -468,26 +468,20 @@ jobs:
         with:
           platform: linux
       - uses: ./.github/actions/install-utoo-npm-baseline
-      - name: Checkout next branch source for utoo-next baseline
+      - name: Build next branch utoo
         run: |
           # The PATH already points at dist/utoo (current branch) — build
           # `next` separately to /tmp so it doesn't shadow. Use a linked
           # git worktree so submodule init works normally inside WORKDIR.
+          # No rust-cache here on purpose: the repo cache budget (10GB) is
+          # already saturated by the per-platform build-pm caches, so adding
+          # a ~400MB utoo-next cache would just LRU-evict more useful ones.
           WORKDIR=/tmp/utoo-next-src
           rm -rf "$WORKDIR"
           git fetch origin next --depth=1
           git worktree add --detach "$WORKDIR" origin/next
-          (cd "$WORKDIR" && git submodule update --init --recursive --depth 1)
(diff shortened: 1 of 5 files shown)


elrrrrrrr:
Splitting bench infra into its own PR (see follow-up). The mimalloc experiment itself is being shelved — bench data showed +7.4% wall on p1_resolve and +708% RSS on p4 vs same-commit system allocator (utoo-next baseline). pm's short-lived CLI / large-object allocation pattern isn't in mimalloc's sweet spot, and resolve-phase gap vs bun (~98%) is algorithmic, not allocator-level.

PR #2868, opened 2026-04-29
ci(pm): add utoo-next baseline to bench-phases-linux

elrrrrrrr:
## Summary

Adds a `utoo-next` baseline to `bench-phases-linux` so PR benches compare against the **exact tip of `next`** rather than only `utoo-npm` (latest published, which can lag merged commits) and `bun` (external).

Why it matters: when investigating perf changes, `utoo-npm` may already include or exclude commits that confound the diff. A same-commit system-allocator baseline is the cleanest control. This was developed while validating a mimalloc experiment (#2867, closed) — the cleaner baseline is the part worth keeping regardless of that result.

## Changes

- `bench/pm-bench-phases.sh`: add `utoo-next` PM target — independent cache dir (`/tmp/utoo-next-bench-cache`), wired through `install_cmd`/`resolve_cmd`/`seed_for_phase`/`write_prepare`/`capture_footprint`, with skip-when-bin-missing logic mirroring the existing `utoo-npm` pattern.
- `.github/workflows/pm-e2e-bench.yml` (`bench-phases-linux`): new `Build next branch utoo` step that uses `git worktree add --detach` (so submodule init works inside the worktree), runs `cargo build --release --bin utoo -p utoo-pm`, exposes `UTOO_NEXT_BIN`. `PM_LIST` becomes `utoo,utoo-next,utoo-npm,bun`. Cache wipe paths extended to include the new cache dir.

## Cost

The next-branch build adds **~3:23** to `bench-phases-linux` (one-time per run, no cache). Total job time on the validation runs: 6:24 → 11:08, well within the PR feedback envelope.

**No `actions/cache` for `target/` on purpose** — the repo's 10GB Actions cache budget is already saturated (~11.9GB / 35 entries when this was authored), and a ~400MB `utoo-next-linux` rust-cache would persistently LRU-evict the per-platform `pm-build-{linux,mac,windows}` caches that the actual `build-*` jobs depend on. Net: trade ~2.5min off bench-phases for cold-rebuilds in build jobs. If the 3:23 ever does become a problem, the right fix is a separate scheduled workflow that uploads a `utoo-next-bin` artifact whenever `next` advances, then bench-phases just `download-artifact` (artifacts have their own retention, don't compete for cache budget).

## Validation

Already exercised on closed PR #2867 — the workflow runs cleanly and produced a 4-PM comparison table (utoo / utoo-next / utoo-npm / bun) on the bench-phases comment.

## Test plan

- [ ] `bench-phases-linux` runs and posts a sticky comment with all 4 PMs.
- [ ] `utoo-next` row appears under each phase (p0/p1/p3/p4) with non-zero wall and RSS.
- [ ] Other CI (build-*, e2e, pm-ci) unaffected.

### DIFF
--- a/.github/workflows/pm-e2e-bench.yml
+++ b/.github/workflows/pm-e2e-bench.yml
@@ -143,6 +143,41 @@ jobs:
           name: utoo-linux-x64
           path: target/x86_64-unknown-linux-gnu/release/utoo
           retention-days: 1
+      # Piggyback on the already-built target/ from the step above: when the
+      # PR is labeled `benchmark`, overlay origin/next's tree onto the current
+      # workdir and re-run cargo build. cargo's incremental compile only
+      # rebuilds crates that actually differ between this PR and next — for
+      # bench-infra-only PRs that's a few seconds (just relink); for PRs that
+      # touch pm sources, only the affected crates rebuild. Either way far
+      # cheaper than a separate cold workspace, and the artifact route avoids
+      # competing with build-* rust-caches for the repo's 10GB cache budget.
+      - name: Build next branch utoo (bench baseline)
+        if: >
+          (github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'benchmark')) ||
+          (github.event_name == 'workflow_dispatch' && inputs.target == 'pm-bench-phases')
+        run: |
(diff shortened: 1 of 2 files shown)


killagu (approved):
LGTM

PR #2869, opened 2026-04-29
perf(pm): preload worker-pool replaces FuturesUnordered

elrrrrrrr:
## Summary

Single-commit re-attempt of [PR #2827](https://github.com/utooland/utoo/pull/2827)'s worker-pool change, freshly cherry-picked on top of latest `next` (which now has aws-lc-rs `#2856`, OnceMap `#2862`, ManifestStore `#2866` already merged).

Replaces the `FuturesUnordered`-based preload with N long-lived `tokio::spawn` workers pulling from a shared `crossbeam_queue::SegQueue<Dep>`. Each worker is independent on tokio's multi-thread runtime, so when one worker is parsing manifest JSON (CPU-bound, simd_json), other workers can still drive their network IO. Termination via `dispatched`/`completed` AtomicUsize + Notify.

Cherry-pick of `7e64c597` from PR #2818, with one resolution: `RegistryClient::fetch_full_manifest` now returns `Arc<FullManifest>` (from #2866) **AND** carries `+ MaybeSend` (from this commit).

## Wasm compat

Native uses real `Send`/`Sync`. Wasm32 uses `MaybeSend`/`MaybeSync` no-op marker traits because `JsFuture` is `!Send` and tasks run via `wasm_bindgen_futures::spawn_local`.

## Local bench (mac, no proxy, ant-design cold-cold, 6 runs)

| binary | median |
|---|---|
| next (origin) | 3.95s |
| **worker-pool (this PR)** | **5.05s** ⚠️ |
| allwins (full PR #2818 stack) | 3.05s |
| bun (npmmirror, --lockfile-only) | 3.00s |

⚠️ **worker-pool alone regresses** vs `next` on local mac (Cloudflare per-IP throttle dominates the per-request CPU savings the architecture buys). Matches PR #2838's driver-hunt finding: worker-pool needs the alloc-cleanup driver bundle to express its win. CI bench-phases on clean GH Actions egress is the authoritative signal — that's where PR #2818 measured −31% wall + 84 vs 55 avg_conc.

This PR is opened as **draft** specifically to get CI bench numbers.

## Test plan

- [x] `cargo build --release -p utoo-pm` clean
- [x] `cargo clippy -p utoo-ruborist -p utoo-pm --all-targets -- -D warnings --no-deps` clean
- [x] `cargo test -p utoo-ruborist` 162 passed, `cargo test -p utoo-pm` 245 passed
- [x] Local cold smoke `utoo deps` on ant-design — exits 0, lockfile generated
[...]

PR #2877, opened 2026-04-30
perf(pm): drop redundant version_list / dist_tags clones on resolve hot path

elrrrrrrr:
## Summary

Two orthogonal alloc-cleanup wins on the dependency-resolve hot path — neither touches behaviour or trait surface. CI bench-phases-linux on the umbrella branch (PR #2818 round-12 vs round-11) measured **−0.13s on p1_resolve with σ tightening from 0.25 → 0.11 (2.3× tighter)**, the biggest single-commit gain of the round-6 → round-12 saga.

4 files changed, +44/−24 lines.

## (1) Drop redundant \`manifest.versions.clone()\`

\`resolve_target_version\` takes \`&[String]\`, so 4 call sites that prefixed it with \`let version_list: Vec<String> = manifest.versions.clone();\` were paying for a pointless deep copy of the entire version list per resolve (50-200 entries on popular packages, ~600 packages on ant-design).

Pass \`&manifest.versions\` / \`&full_manifest.versions\` directly. Saves ~80 String allocs × ~600 packages = ~48k allocs per ant-design install.

Sites:
- \`crates/ruborist/src/traits/registry.rs\` — default \`fetch_version_manifest\` + default \`resolve_package\` non-semver branch
- \`crates/ruborist/src/resolver/registry.rs\` — \`resolve_from_manifest\`
- \`crates/pm/src/cmd/view.rs\` — \`utoo view\` command

## (2) \`mem::take\` version_list / dist_tags into VersionsInfo

The 200 path of \`resolve_full_manifest\` previously deep-cloned \`manifest.versions\` (Vec<String>) and \`manifest.dist_tags\` (HashMap) per package fetch to fill the \`versions_info\` cache slot, while the same fields also moved into \`Arc<FullManifest>\` for the \`full_manifests\` slot — two owners holding identical copies of the same data.

Use \`std::mem::take\` to MOVE both fields out of \`manifest\` directly into the new \`VersionsInfo\`. The cached \`Arc<FullManifest>\` keeps \`raw\` + scalar metadata; the canonical \`version_list\` / \`dist_tags\` now live in the \`versions_info\` cache slot only. \`resolve_via_full_manifest\`'s Full branch reads them via \`cache.get_versions(name)\` (same code path that filled the slot, atomic w.r.t. the inflight gate).

\`get_core_version(&resolved)\` still works on the cached \`Arc<FullManifest>\` because it parses the single version on demand from \`raw\` (preserved) — not from the moved-out \`versions\` field. Saves another ~80 String + 1 HashMap clone × ~600 packages per install.

## Why this is more than alloc-cleanup

The \`versions.clone()\` / \`dist_tags.clone()\` calls run inside \`resolve_full_manifest\`'s \`inflight_full.get_or_init\` worker closure on the **main task / aggregator task** — not in the per-package resolver spawned futures. So even with worker-pool architecture (PR #2869), this clone bottleneck is sequential. \`mem::take\` directly removes that sequential work, which is why bench shows real wall-time savings instead of the noise-bound result alloc-cleanups normally produce in a parallel-resolve baseline.

## Bench evidence (CI Linux ant-design / npmjs)

| commit | indicatif | short-circuit | **alloc cleanup** | p1_resolve | σ |
|---|---|---|---|---|---|
| round-10 (\`f122478b\`) | TTY-gated | ❌ | ❌ | 2.58s | 0.07 |
| round-11 (\`5ca65e70\`) | TTY-gated | ✅ | ❌ | 2.59s | 0.25 |
| **round-12 (\`b2498e91\`, this PR's content)** | **TTY-gated** | **✅** | **✅** | **2.46s** | **0.11** |

vs \`utoo-next\` baseline (5.19s ±0.14): −2.73s (−53%).

## Test plan

- [x] \`cargo build --release -p utoo-pm\` clean
- [x] \`cargo clippy -p utoo-ruborist -p utoo-pm --all-targets -- -D warnings --no-deps\` clean
- [x] \`cargo test -p utoo-ruborist\`: 162 + 10 doctests pass
- [x] CI \`pm-bench-phases-linux\` on umbrella PR #2818 round-12 (\`b2498e91\`): p1=2.46s ±0.11 (commit ID \`7b40566\`)
- [ ] CI \`pm-bench-phases-linux\` on this PR (clean against \`next\`) — expecting independent reproduction

### DIFF
--- a/crates/pm/src/cmd/view.rs
+++ b/crates/pm/src/cmd/view.rs
@@ -27,10 +27,9 @@ pub async fn view(package_spec: &str) -> Result<()> {
     tracing::debug!("Fetched package info: {full_manifest:?}");
 
     // Resolve version and get full VersionManifest (with all display fields)
-    let version_list: Vec<String> = full_manifest.versions.clone();
     let resolved_version = utoo_ruborist::resolver::version::resolve_target_version(
         &full_manifest.dist_tags,
-        &version_list,
+        &full_manifest.versions,
         version_spec,
     )
     .map_err(|e| {
--- a/crates/ruborist/src/resolver/registry.rs
+++ b/crates/ruborist/src/resolver/registry.rs
@@ -137,14 +137,12 @@ pub fn resolve_from_manifest<E: std::error::Error + 'static>(
     manifest: &FullManifest,
     spec: &str,
 ) -> Result<ResolvedPackage, ResolveError<E>> {
-    let version_list: Vec<String> = manifest.versions.clone();
-
-    if version_list.is_empty() {
+    if manifest.versions.is_empty() {
         return Err(ResolveError::NoVersions(manifest.name.clone()));
     }
 
     // Resolve version using shared logic
-    let resolved_version = resolve_target_version(&manifest.dist_tags, &version_list, spec)
+    let resolved_version = resolve_target_version(&manifest.dist_tags, &manifest.versions, spec)
         .map_err(|e| ResolveError::Version(format!("{}@{}: {}", manifest.name, spec, e)))?;
 
     // Get manifest for resolved version (lazy: parse from Value on demand)
--- a/crates/ruborist/src/service/registry.rs
+++ b/crates/ruborist/src/service/registry.rs
@@ -244,11 +244,19 @@ impl UnifiedRegistry {
                 .await
                 .map_err(RegistryError)?
                 {
-                    manifest::FetchManifestResult::Ok(manifest, new_etag) => {
+                    manifest::FetchManifestResult::Ok(mut manifest, new_etag) => {
+                        // MOVE version_list / dist_tags out of `manifest`
+                        // into `VersionsInfo` instead of cloning. The cached
+                        // `Arc<FullManifest>` keeps `raw` + scalar metadata;
+                        // canonical version_list / dist_tags now live in the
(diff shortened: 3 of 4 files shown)


elrrrrrrr:
Closing per Gemini's review — the `mem::take` approach breaks two public API contracts:

1. **`pub fn resolve_from_manifest(&FullManifest, ...)`** silently returns `NoVersions` if the caller hands us a cache-returned manifest (now empty `versions`).
2. **`RegistryClient::fetch_full_manifest()`** trait contract is violated — its returned `Arc<FullManifest>` has empty `versions` / `dist_tags` fields, breaking any external consumer and the default `fetch_versions_info` impl that calls `self.fetch_full_manifest().clone()`.

Both are real BC breaks, not just internal refactors.

Will re-open in two cleaner forms:

- **(A)** strip to just the 4-site `let version_list = manifest.versions.clone()` removal — strictly correct, zero API impact, modest gain.
- **(C)** lift `FullManifest::versions` to `Arc<Vec<String>>` (and `dist_tags` to `Arc<HashMap<…>>`) so cloning is a ref-count bump — preserves the API contract while keeping the −0.13s gain that came from the `mem::take` saving.

Bench evidence (PR #2818 round-12, p1=2.46s ±0.11 vs round-11 p1=2.59s ±0.25) carries forward.

PR #2878, opened 2026-04-30
perf(pm): introduce VersionsRef<'a> view for resolve_target_version

elrrrrrrr:
## Summary

Refactor the version-resolution read path to take a borrowed view (`VersionsRef<'a>`) instead of two separate `&HashMap<…>` / `&[String]` arguments. The resolution logic only needs read access to a versions list + dist-tag map; everything else on a `FullManifest` (raw bytes, time, maintainers, …) is irrelevant.

```rust
pub struct VersionsRef<'a> {
    pub versions: &'a [String],
    pub dist_tags: &'a HashMap<String, String>,
}

impl<'a> From<&'a FullManifest> for VersionsRef<'a> { ... }
impl<'a> From<&'a VersionsInfo> for VersionsRef<'a> { ... }
impl<'a> From<&'a Versions> for VersionsRef<'a> { ... }
```

`resolve_target_version` now takes `VersionsRef<'_>`. Callers construct via `From` based on whichever in-memory shape they have:

```rust
// 200-path: read directly from FullManifest
let resolved = resolve_target_version((&*full).into(), spec)?;

// 304-path: read from cached VersionsInfo
let resolved = resolve_target_version((&*info).into(), spec)?;
```

8 files changed, +98/−42 lines.

## Why this design (replaces the closed PR #2877)

The prior `mem::take`-based approach (PR #2877) tried to dodge the deep clones at `resolve_full_manifest`'s 200 path by moving `versions` / `dist_tags` out of the cached `Arc<FullManifest>` into the `VersionsInfo` slot. Gemini correctly identified that this broke two public API contracts:

- `pub fn resolve_from_manifest(&FullManifest, …)` would silently return `NoVersions` if handed a cache-returned manifest.
- `RegistryClient::fetch_full_manifest` trait contract is violated — its returned `Arc<FullManifest>` would have empty fields, breaking external consumers and the default `fetch_versions_info` impl.

The view abstraction sidesteps both problems:

- `FullManifest::versions` / `dist_tags` stay as owned `Vec` / `HashMap` — fields remain populated, no API break, no `mem::take`.
- The duplicate-storage bug (cache holds versions+dist_tags in BOTH `full_manifests` and `versions_info` slots, requiring a deep clone to fill the second) is now isolated to the *write* side. A follow-up PR can remove the duplicate fill on the 200 path; readers already dispatch via the view, so swapping the source is source-compat for them.

## Why `VersionsRef<'a>` (not `'static` or `Arc<…>`)

| Approach | Hot path cost | API impact | Issue |
|---|---|---|---|
| `'static` references | zero | API rigid | Forces `Box::leak` (memory leak on every fetch) or `LazyLock` (single global) |
| `Arc<Vec<String>>` field | atomic ref-bump per resolve | type signature widens | OK but ergonomics tax forever |
| **`VersionsRef<'a>`** | **zero (borrow)** | **new type, callers adapt** | **None** |

Lifetime parameter is the idiomatic Rust solution: zero-cost, borrow-checker enforces correctness, and the lifetime never escapes a single function call (no propagation to long-lived structs). Most callers won't even need to write the `'a` thanks to elision.

## Drop redundant `manifest.versions.clone()` along the way

Four call sites had `let version_list: Vec<String> = manifest.versions.clone();` just to feed the old `&[String]` parameter — now redundant since `VersionsRef::from` borrows. Dropped at:

- `traits/registry.rs` — default `fetch_version_manifest` + non-semver `resolve_package`
- `resolver/registry.rs::resolve_from_manifest`
- `pm/cmd/view.rs`

Saves ~80 String allocs × ~600 packages = ~48k allocs per ant-design install. Worker-pool baseline parallelises this work so isolated bench signal will be small (within σ); the architectural cleanup is the main value here.

## Naming choice

Considered `VersionsView` first but rejected — that conflicts with the `*View` family in `model/package_json.rs` (`DepsView`, `IdentityView`, `PackageInstallView`, `ScriptsView`, `EnginesView`) which are owned schema-deserialization subsets. `VersionsRef` follows Rust's standard borrowed/owned naming pair (`OsStr`/`OsString`, `Path`/`PathBuf`, `str`/`String`) — pairs naturally with the existing owned `Versions` struct in `service/cache.rs`.

## Test plan

- [x] `cargo clippy -p utoo-ruborist -p utoo-pm --all-targets -- -D warnings --no-deps` clean
- [x] `cargo test -p utoo-ruborist`: 162 + 10 doctests pass
- [x] `cargo test -p utoo-pm`: 245 + 10 doctests pass
- [ ] CI `pm-bench-phases-linux` — expecting neutral on p1_resolve (within σ); lays groundwork for the follow-up that consolidates the duplicate-storage on the 200 path.

### DIFF
--- a/crates/pm/src/cmd/view.rs
+++ b/crates/pm/src/cmd/view.rs
@@ -27,10 +27,8 @@ pub async fn view(package_spec: &str) -> Result<()> {
     tracing::debug!("Fetched package info: {full_manifest:?}");
 
     // Resolve version and get full VersionManifest (with all display fields)
-    let version_list: Vec<String> = full_manifest.versions.clone();
     let resolved_version = utoo_ruborist::resolver::version::resolve_target_version(
-        &full_manifest.dist_tags,
-        &version_list,
+        (&full_manifest).into(),
         version_spec,
     )
     .map_err(|e| {
--- a/crates/ruborist/src/lib.rs
+++ b/crates/ruborist/src/lib.rs
@@ -41,7 +41,9 @@ pub mod graph {
 
 /// Package manifest types.
 pub mod manifest {
-    pub use crate::model::manifest::{CoreVersionManifest, Dist, FullManifest, VersionManifest};
+    pub use crate::model::manifest::{
+        CoreVersionManifest, Dist, FullManifest, VersionManifest, VersionsRef,
+    };
     pub use crate::model::package_json::{
         DepsView, EnginesView, IdentityView, PackageInstallView, PackageJson, PublishConfig,
         ScriptsView,
--- a/crates/ruborist/src/model/manifest.rs
+++ b/crates/ruborist/src/model/manifest.rs
@@ -9,6 +9,44 @@ use std::sync::Arc;
 use serde::{Deserialize, Deserializer, Serialize};
 use serde_json::Value;
 
+/// Borrowed view of the data needed to resolve a version spec — a slice of
+/// available versions plus a dist-tag map.
+///
+/// The version-resolution logic ([`crate::resolver::version::resolve_target_version`])
+/// only needs read access to these two pieces of data; everything else on a
+/// `FullManifest` (raw bytes, time map, maintainers, …) is irrelevant. By
+/// borrowing them through a unified view we can serve the same resolver from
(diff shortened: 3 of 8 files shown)


killagu (approved):
LGTM

PR #2874, opened 2026-04-30
perf(pm): manifest resolve hot-path — rayon offload + network layer (h1+pool+dns)

elrrrrrrr:
## Summary

Manifest resolve hot-path overhaul on rayon-only architecture (no worker-pool refactor), plus network-layer optimizations cherry-picked from PR #2818. After noise-adjusted measurement, this stack reaches within ~0.2s of the worker-pool baseline (#2818) on Linux ant-design / npmjs.

| Commit | Change |
|---|---|
| `2b54a57d` | rayon offload for `simd_json::serde::from_slice` (fetch path) + `extract_version` reparse |
| `b43ef5ce` | short-circuit `cache.get_version_manifest(name, resolved)` before rayon reparse |
| `eb95b56d` | preload concurrency 64 → 128 |
| `7f7f08b9` | HTTP/1.1 + connection pool for manifest fetching (cherry-pick from PR #2818) |
| `440c501a` | round-robin DNS addresses across connection pool (cherry-pick from PR #2818) |

## Why network-layer optimizations were added

Initial validation against worker-pool (#2818, p1=2.70s) showed rayon-only stuck at p1=3.6–4.8s — a ~1.0s gap that PR #2872's ablation revealed was **mostly network-layer, not architectural**:

| Variant (worker-pool path) | p1_resolve | Δ vs PR #2818 |
|---|---|---|
| #2818 full (h1+pool+dns) | 2.70s | — |
| #2872 (no dns) | 3.07s | +0.37s |
| revert h1+pool | 7.34s (σ 4.82, max 12.9s) | +4.64s ❌ |
| revert dns rr | 3.36s (σ 0.67, max 4.13s) | +0.66s ❌ |

reqwest's default H2 multiplexing causes head-of-line blocking on Cloudflare-fronted npmjs (1 slow tail response stalls the whole phase). A forced HTTP/1.1 pool with DNS round-robin across the multi-IP CDN restores parallelism.

## CI bench-phases-linux (Linux ant-design / npmjs, noise-adjusted)

Two samples on a slow-runner day; ratio normalization against bun in the same run isolates the runner-induced shift.

| Config | utoo p1 (raw) | bun p1 (raw) | utoo/bun | Normalized to bun typical (~2.0s) |
|---|---|---|---|---|
| **rayon + h1+pool+dns (this PR)** | 4.21s median | 2.99s | **1.40** | **~2.80s** |
| worker-pool full (#2818) | 2.70s | 2.06s | 1.31 | ~2.62s |
[...]

PR #2882, opened 2026-05-01
perf(pm): dedup version-manifest disk writes via OnceMap

elrrrrrrr:
## Problem

`DiskManifestStore::store_version_manifest` is called from `service/registry.rs:388` inside the `inflight_version<(name, original_spec)>` gate. The gate keys on the **original spec**, but the disk write keys on the **resolved version**. Two sibling specs (`^4.17.0`, `~4.17.20`) resolving to the same `4.17.21` each pass through different gates and both spawn `write_json` for the same path:

```text
spec="^4.17.0"  ──┬─ inflight gate 1 ─→ resolved 4.17.21 ─→ tokio::spawn write 4.17.21.json
spec="~4.17.20" ──┴─ inflight gate 2 ─→ resolved 4.17.21 ─→ tokio::spawn write 4.17.21.json
spec="4.x"      ────  inflight gate 3 ─→ resolved 4.17.21 ─→ tokio::spawn write 4.17.21.json
```

Each duplicate does:
- redundant `serde_json::to_vec` (~5 KB × N)
- redundant disk write
- truncate-race window where a concurrent reader sees partial JSON (read fails → cache miss)

`store_versions` (versions.json) is safe — it sits inside `inflight_full<name>` whose key matches the write key.

## Fix

Add an `OnceMap<(name, resolved_version), ()>` to `DiskManifestStore`. Reuse the existing `OnceMap::register` (sync, atomic "first wins") that ruborist already uses for `inflight_full` / `inflight_version`. Subsequent callers for the same `(name, version)` get `None` from `register` and skip the spawn entirely.

```rust
let Some(notify) = self.inflight_version_writes.register(key.clone()) else {
    return;  // duplicate, skip
};
let inflight = Arc::clone(&self.inflight_version_writes);
tokio::spawn(async move {
    write_json(&path, &*manifest).await;
    inflight.complete(key, Some(()), notify);
});
```

## Impact

ant-design preload: ~600 specs → ~250 unique versions → ~350 redundant `spawn + serialize + write` cycles eliminated, ~1.75 MB redundant disk write avoided. Wall-clock impact within σ (writes are fire-and-forget off the resolve hot path), but reduces CPU/IO contention with preload workers and closes the truncate-race window.

## Test plan

- [x] `cargo build --profile release-local -p utoo-pm` clean
- [x] `cargo clippy -p utoo-pm -p utoo-ruborist --all-targets -- -D warnings --no-deps` clean
- [x] `cargo test -p utoo-pm` — 245 passed, 0 failed
- [ ] CI \`bench-phases-linux\` — confirm no regression

### DIFF
--- a/crates/pm/src/util/manifest_store.rs
+++ b/crates/pm/src/util/manifest_store.rs
@@ -16,16 +16,27 @@ use async_trait::async_trait;
 use serde::Serialize;
 use utoo_ruborist::model::manifest::CoreVersionManifest;
 use utoo_ruborist::service::{ManifestStore, VersionsInfo};
+use utoo_ruborist::util::OnceMap;
 
 use crate::util::json::read_json_file;
 
 pub struct DiskManifestStore {
     cache_dir: PathBuf,
+    /// Dedup `(name, resolved_version)` disk writes. The upstream
+    /// `inflight_version<(name, original_spec)>` gate keys on the original
+    /// spec, so two specs (`^4.17.0`, `~4.17.20`) resolving to the same
+    /// `4.17.21` each call `store_version_manifest` and would otherwise
+    /// spawn duplicate `write_json` tasks racing on truncate of the same
+    /// path. `OnceMap::register` is a sync, atomic "first wins" check.
+    inflight_version_writes: Arc<OnceMap<(String, String), ()>>,
 }
 
 impl DiskManifestStore {
     pub fn new(cache_dir: PathBuf) -> Self {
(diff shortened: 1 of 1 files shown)


elrrrrrrr:
Superseded by #2874 commit `f96c6e82` (`perf(ruborist): dedup store_version_manifest disk writes per resolved version`), which is identical in scope: `DashSet<(name, resolved_version)>` gating around the disk write. Closing to consolidate.

PR #2884, opened 2026-05-02
perf(pm): bin-link sync std::fs (drop tokio fs hop, sort + dedupe)

elrrrrrrr:
## Summary
- Bin-link hot path was async `tokio::fs::*`; each call routed through `spawn_blocking` + mpsc + thread-hop, dominating the actual syscall cost.
- Switch to sync `std::fs::*` on a deterministic sorted + deduped queue.

## Benchmark — ant-design, linux ubuntu-latest, 3 runs median (4-variant harness, previous PR commit)

| phase           | serial-async | **serial-sync** | parallel-tokio | parallel-rayon |
|-----------------|-------------:|----------------:|---------------:|---------------:|
| p0_full_cold    |        38 ms |        **6 ms** |           8 ms |           3 ms |
| p3_cold_install |        36 ms |        **6 ms** |           8 ms |           3 ms |
| p4_warm_link    |        36 ms |        **6 ms** |           7 ms |           4 ms |

228 bins linked per run. Headline: **30 ms (≈83%) saved** by going async → sync. Adding parallelism on top recovered at most an additional 3 ms, well within stddev (~1 ms), so not shipped — keeps the diff a single `for` loop with no concurrency primitives.

`parallel-tokio` was actually slower than `serial-sync` (7 ms vs 6 ms), confirming the `spawn_blocking` overhead theory rather than CPU contention.

## Bin-name collision strategy
Mirrors npm `bin-links/lib/link-gently.js`:
- sort packages by path so `.bin/<name>` collisions (e.g. `svgo` vs `svgo-browser` in ant-design) resolve deterministically (shallower / lex-smaller path wins).
- `HashSet<PathBuf>` dedupe on `link_path` — first writer wins, subsequent same-name jobs dropped silently.

## Test plan
- [x] `cargo build -p utoo-pm`
- [x] `cargo clippy -p utoo-pm --all-targets --no-deps -- -D warnings`
- [x] `cargo test -p utoo-pm service::package` (16/16) + `util::linker` (6/6)
- [x] CI bench (run 25254083182) confirmed serial-sync = 6 ms p4_warm_link median

## Follow-ups (not in this PR)
- `linkat`/`symlinkat` + `openat`-cached parent fd: theoretical extra 1–4 ms; meaningful at the new 6 ms scale, orthogonal to this change.
- The `[bin-link]` telemetry line stays; can be dropped later if we don't want it in steady-state install output.

### DIFF
--- a/crates/pm/src/service/package.rs
+++ b/crates/pm/src/service/package.rs
@@ -282,65 +282,95 @@ impl PackageService {
     /// Queue contains (PackageInfo, is_optional) tuples - is_optional is not used
     /// here as binary linking happens only for successfully installed packages.
     ///
-    /// Emits a single `[bin-link]` line on stderr summarising the batch
-    /// (count + wall time). Captured by the bench harness via hyperfine's
-    /// `--show-output` so we can quantify what fraction of warm-link wall
-    /// is bin linking before deciding whether to parallelise.
+    /// Strategy:
+    ///   1. Sort packages by path so module-local bin-name collisions resolve
+    ///      deterministically (shallower / lex-smaller path wins).
+    ///   2. Dedupe by `link_path` in a `HashSet` — first writer wins, mirroring
+    ///      npm `bin-links/lib/link-gently.js`'s seen-set. Without this,
+    ///      multiple packages declaring the same bin (e.g. `svgo` + `svgo-browser`)
+    ///      both call into `link_bin`, which resolves via "remove existing,
+    ///      symlink new" and yields non-deterministic last-writer-wins.
+    ///   3. Walk the deduped jobs serially calling sync `link_bin`. Each
+    ///      `link_bin` is `try_exists` + `symlink_metadata` + `symlink(2)` —
+    ///      microsecond-class syscalls where async `tokio::fs::*` overhead
+    ///      (spawn_blocking + mpsc + thread hop) dominates real cost. Bench
+    ///      data on ant-design (228 bins) showed 36ms async → 6ms sync; adding
+    ///      try_join_all/rayon parallelism only saved an additional 2-3ms over
+    ///      sync, well within stddev — not worth the concurrency complexity.
+    ///
+    /// Emits one `[bin-link]` line on stdout summarising the batch (captured
+    /// by `bench/pm-bench-phases.sh` via hyperfine `--show-output`).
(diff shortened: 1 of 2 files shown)


elrrrrrrr:
## Bin-link perf investigation summary

Used a 4-variant in-tree harness (intermediate commit `21f0bc79`, since replaced) to compare implementations on the same deduped job set, wiping `node_modules/.bin` between runs for apples-to-apples.

**ant-design / linux ubuntu-latest / 3 runs median / 228 bins**

| phase           | serial-async | **serial-sync** | parallel-tokio | parallel-rayon |
|-----------------|-------------:|----------------:|---------------:|---------------:|
| p0_full_cold    |        38 ms |        **6 ms** |           8 ms |           3 ms |
| p3_cold_install |        36 ms |        **6 ms** |           8 ms |           3 ms |
| p4_warm_link    |        36 ms |        **6 ms** |           7 ms |           4 ms |

**Key finding**: The 30 ms (≈83%) win is purely from dropping the `tokio::fs::*` hop — every bin-link call was paying spawn_blocking + mpsc + thread-hop overhead for what amounts to a few microsecond-class syscalls. `parallel-tokio` (7 ms) was actually slower than `serial-sync` (6 ms), confirming spawn_blocking is the dominant cost rather than CPU contention.

**Why not also parallelise**: `parallel-rayon` (4 ms) saves ≈2 ms more than `serial-sync` — within the noise band (stddev ~1 ms) and below the 20 ms threshold for shipping additional complexity. A single sync `for` loop is simpler than `rayon::spawn` + `into_par_iter` + `oneshot` channel.

**EEXIST race / dedupe**: With a serial loop the race window is gone, but kept the `sort + HashSet` dedupe step anyway because it's how npm's `bin-links/lib/link-gently.js` resolves the `svgo` vs `svgo-browser` style collision deterministically (shallower path wins, first writer claims the slot, subsequent same-name jobs are dropped silently).

**Follow-up worth investigating** (not in this PR): `openat` + `symlinkat` to cache the `.bin` parent fd — theoretical extra 1–4 ms by skipping per-call dentry walk. More meaningful at the new 6 ms scale than at 36 ms. Orthogonal to this change.

killagu (approved):
LGTM

PR #2888, opened 2026-05-03
chore(pm): drop [bin-link] stdout summary

elrrrrrrr:
## Summary

- The `println!(\"[bin-link] ...\")` added in #2884 hits stdout on every `ut install`, but nothing actually consumes it — `bench/pm-bench-phases.sh` enables hyperfine `--show-output` but doesn't grep for the line.
- Remove the print, the wall-time/total/processed/skipped/deduped counters that only fed it, and the doc comment claim that bench captures it.
- Sort + dedupe + sync `link_bin` core logic is unchanged.

## Test plan

- [x] `cargo fmt`
- [x] `cargo clippy -p utoo-pm --all-targets -- -D warnings --no-deps`
- [ ] `ut install` on a project with bins — stdout no longer shows `[bin-link] ...`


killagu (approved):
LGTM

PR #2887, opened 2026-05-03
perf(pm): install hot-path micro-opts (OnceMap API + cache-hit fast-paths, file_type, binary mirror gate)

elrrrrrrr:
## Summary

Tiny, structurally-clean install hot-path optimizations. **4 files, +42/−3 lines.**

| Change | Files |
|---|---|
| **`OnceMap::is_done` + `get` accessors** — sync, single-`DashMap::get` probes for completed entries | `crates/ruborist/src/util/oncemap.rs` |
| **`download_to_cache` cache-hit fast-path** — uses `OnceMap::get` to skip per-call `format!` / `to_string` allocs + future construction when the key is `Done` | `crates/pm/src/util/downloader.rs` |
| **`clone_package_once` cache-hit fast-path** — uses `OnceMap::is_done` to skip 5 string allocs + future construction when the target is already cloned | `crates/pm/src/util/cloner.rs` |
| **`find_real_src` uses `file_type()` over `metadata()`** — `d_type` from cached `readdir` skips an `lstat` per entry on ext4/btrfs/xfs (~3-5K stat syscalls saved on cold install) | `crates/pm/src/util/cloner.rs` |
| **Skip `update_package_binary` for npm.org** — early-return when `should_skip_binary_mirror()`; avoids per-process `binary-mirror-config/latest` fetch + per-package China mirrors hashmap lookup | `crates/pm/src/service/binary.rs` |

## Why so small

This branch was originally a longer stack with two more buckets that got separated out after CI investigation:

1. **Install streaming triplet** (TTY-gate `progress_inc/set_length` + `install_packages` `FuturesUnordered` layer drain + zero-copy tar entries via `data_offset/data_len` slicing). Reverted because A/B testing showed it widened utoo `p0_full_cold` σ to ~14× baseline. The streaming refactor genuinely tightens p3 σ from ~1.5s to 0.07s and prevents long-tail tarball stalls — but it relocates timing variance from p3 onto p0, which needs more design (e.g. layered semaphore for long-tail handling). Parked on `experiment/install-streaming-triplet`.
2. **Permit-drop before extract** (release `DOWNLOAD_SEMAPHORE` permit between download and extract). Reverted as part of the σ investigation, found innocent. Not currently scoped for follow-up.
[...]

PR #2902, opened 2026-05-06
experiment(pm): install hot-path streaming triplet (TTY-gate + FuturesUnordered + zero-copy)

elrrrrrrr:
## What this PR is for

Observation experiment, not a merge candidate. Companion to #2887.

#2887 ships the 5 small structurally-clean micro-opts. This PR carries the bigger triplet that got A/B-reverted out of #2887 for widening utoo's p0_full_cold σ to ~14× baseline. We want to keep tracking it because the **mean** is fine and the **p3** behavior is meaningfully better with it active — the issue is purely σ-shape, and that needs a fix (probably a layered semaphore for long-tail tarballs) before it's mergeable.

## What's in the triplet (single commit `b92fc85a`)

1. **TTY-gate `progress_inc` / `progress_set_length` helpers** — non-TTY callers (CI, piped output) skip indicatif's `Mutex<ProgressState>` write-lock. ~9k Mutex acquisitions removed per CI install.
2. **`install_packages` per-layer `FuturesUnordered` streaming** — drains tasks in completion order so a slow tarball doesn't pin its siblings; surfaces the first error early.
3. **Zero-copy tar entries** — `ExtractedEntry { content: Vec<u8> }` → `(data_offset, data_len)` slicing into the shared decompressed buffer. Eliminates per-file `Vec<u8>` alloc + copy.

## Tradeoff observed in earlier CI A/B (#2887 σ investigation)

| State | utoo `p0_full_cold` σ ratio (vs utoo-next) | utoo `p3_cold_install` σ |
|---|---|---|
| Triplet ACTIVE | **11–31×** baseline (wide) | **0.07–0.5s** (very tight) |
| Triplet REVERTED (#2887 ships state) | 0.3–1.1× (parity) | 0.5–14s (occasional 39s long-tail outlier) |

The streaming refactor IS doing its job — preventing one long-tail tarball from pinning the whole layer. But by widening the per-task latency distribution, p0's wall variance climbs.

## What we want to observe

- p0 σ behavior across multiple bench-phases-linux runs
- p3 σ tightness — should be visibly better than #2887's
- Whether p3 occasional outliers (39s) really go away with streaming active
- iCtx / vCtx delta — streaming changes the spawn_blocking dispatch pattern

## Future work (not in this PR)

A layered approach to long-tail handling:
- Detect long-tail tarballs (size estimate from gzip footer, or measured download time)
- Route them through a smaller secondary semaphore so they can't starve install layer's slot turnover
- Or: split the install loop's drain into "fast" (cache hit / small) and "slow" (long-tail) lanes, drain fast in completion order but slow in spawn order

This PR is just the data-gathering vehicle. No merge intent until the design is sound.

## Companion

- #2887 — the 5 small micro-opts that did ship


elrrrrrrr:
Superseded by #2924 / #2929 stack — experiment archived.

PR #2903, opened 2026-05-07
perf(pm): TTY-gate progress bar on install hot-path

elrrrrrrr:
## Context

Split 1/3 of the install hot-path triplet from #2902. Each piece is shipped as an independent draft so we can A/B them against the new `next` baseline (#2887 just merged) without confounds.

## What this PR does

`install.rs` had 8 raw `PROGRESS_BAR.inc(1)` calls plus a `set_length` that bypassed the `IS_TTY` short-circuit already present in `ProgressReceiver`. indicatif always takes the internal `Mutex<ProgressState>` write-lock even with a hidden draw target, so non-TTY runs (CI, piped output) were paying ~9k Mutex acquisitions per install.

Wrapped them in `progress_inc` / `progress_set_length` helpers that match the receiver's gating pattern.

## A/B target

CI `bench-phases-linux` should show:
- p3_cold_install: should be ≈ neutral or slightly faster (less mutex contention in install hot loop)
- p0_full_cold: should be neutral or slightly faster — does NOT add new concurrency, only removes work
- vCtx: small drop from removed mutex acquisitions

If p0 wall regresses on this alone, the TTY-gate is implicated. If it doesn't, the regression observed in #2902 is from the streaming/zero-copy combo.

## Companion experiments

- #2902 — full triplet (TTY-gate + streaming + zero-copy), p0 wall regresses
- experiment/install-streaming-drain — FuturesUnordered drain in isolation
- experiment/install-zero-copy-tar — zero-copy tar entries in isolation

No merge intent — observation experiment.

### DIFF
--- a/.github/workflows/pm-e2e-bench.yml
+++ b/.github/workflows/pm-e2e-bench.yml
@@ -154,7 +154,7 @@ jobs:
       - name: Build next branch utoo (bench baseline)
         if: >
           (github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'benchmark')) ||
-          (github.event_name == 'workflow_dispatch' && inputs.target == 'pm-bench-phases')
+          (github.event_name == 'workflow_dispatch' && (inputs.target == 'pm-bench-phases' || inputs.target == 'pm-bench-pcap'))
         run: |
           # The previous Upload step has already shipped the PR binary; we
           # can now overwrite target/.../release/utoo safely.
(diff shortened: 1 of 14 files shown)


elrrrrrrr:
Closing — superseded by #2911. The TTY-gate change is preserved there, with the +1.75s p3 regression observed on this branch fixed structurally by partitioning install_packages into classify + 3-phase pipeline (the synchronous spawn-loop burst that caused tokio runtime starvation is gone).

PR #2904, opened 2026-05-07
perf(pm): stream install_packages layer drain via FuturesUnordered

elrrrrrrr:
## Context

Split 2/3 of the install hot-path triplet from #2902. Each piece is shipped as an independent draft so we can A/B them against the new `next` baseline (#2887 just merged) without confounds.

## What this PR does

Replaced `for task in clone_tasks { task.await?? }` with `FuturesUnordered`. The previous serial join walked tasks in spawn order, so a single slow tarball blocked the layer's `await` loop while later finishers piled up unobserved — pure idle latency between layers on deep trees.

With FuturesUnordered the layer drains in completion order and propagates the first error early.

## A/B target

CI \`bench-phases-linux\` should show:
- p3_cold_install: should be visibly tighter σ (long-tail tarballs no longer block sibling completions)
- p0_full_cold: this is the suspect change — by widening peak install concurrency it may steal scheduler/IO bandwidth from the resolve+download path that p0 runs concurrently. Watch wall + vCtx.
- iCtx delta: streaming changes the spawn_blocking dispatch pattern

If p0 wall regresses on this alone (and TTY-gate / zero-copy in their own PRs don't), the streaming change is the regression source — and the layered-semaphore fix proposed in #2902 is the path forward.

## Companion experiments

- #2902 — full triplet (TTY-gate + streaming + zero-copy)
- #2903 — TTY-gate alone
- experiment/install-zero-copy-tar — zero-copy tar entries alone

No merge intent — observation experiment.

### DIFF
--- a/.github/workflows/pm-e2e-bench.yml
+++ b/.github/workflows/pm-e2e-bench.yml
@@ -154,7 +154,7 @@ jobs:
       - name: Build next branch utoo (bench baseline)
         if: >
           (github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'benchmark')) ||
-          (github.event_name == 'workflow_dispatch' && inputs.target == 'pm-bench-phases')
+          (github.event_name == 'workflow_dispatch' && (inputs.target == 'pm-bench-phases' || inputs.target == 'pm-bench-pcap'))
         run: |
           # The previous Upload step has already shipped the PR binary; we
           # can now overwrite target/.../release/utoo safely.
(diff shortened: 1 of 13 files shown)


elrrrrrrr:
Closing — N=2 phases bench on the new architecture (post-#2905 zero-copy + par_chunks merge) shows streaming drain is exactly neutral: p0 Δ ≈ 0, p3 Δ ≈ 0, σ all healthy. Without a measurable benefit there's no reason to add the surface area. Filing the design observation as a memory; the FuturesUnordered drain pattern remains a valid tool if a future change exposes long-tail tarball latency on a single layer.

PR #2905, opened 2026-05-07
perf(pm): zero-copy tar entries in extract_tarball_sync

elrrrrrrr:
## What

Two compounding changes to `extract_tarball_sync`:

1. **Zero-copy tar entries**. `ExtractedEntry` previously held an owned `Vec<u8>` copied via `read_to_end` for every file. Each tarball with N files paid N allocs + N memcpys before the parallel-write phase even started. Switched to `(data_offset, data_len)` indices into the decompressed buffer; the rayon write phase slices into the shared `&[u8]` directly. Tar Slip + path-traversal guards untouched, plus a buffer-overrun check on the recorded offsets.

2. **Chunked parallel writes** (`par_chunks(64)` instead of `par_iter`). On its own, zero-copy made the per-file write phase too tight: `par_iter` then spread every individual write across all rayon workers, queuing N concurrent IO syscalls into the kernel scheduler at burst. On a constrained disk (GHA Azure) the queue depth blew up — pcap+iostat showed `util_max=92%`, `w_await` peaks of `490ms`, paired with TCP-level `retx=123` from blocked write threads stalling the tokio runtime's socket draining. `par_chunks(64)` keeps cross-thread parallelism (each rayon worker takes a 64-file chunk in parallel with sibling chunks) but bounds in-flight write count per worker to 1.

## Why this combination

The two changes were debugged in isolation across `experiment/zero-copy-iter-write`, `experiment/zero-copy-par-chunks-{32,64,128}`, and `experiment/zero-copy-rayon-cap`. Picking `64` as the chunk size:

| chunk | retx | zwin | util_max | w_await |
|---|---|---|---|---|
| `1` (par_iter) | 123 | 9 | 92% | 490ms |
| `32` | 0 | 0 | 105% | 268ms |
| **`64` (this PR)** | **10** | **1** | **81%** | **160ms** |
| `128` | 11 | **16** ← back-pressure starts | 90% | 84ms |
| `∞` (iter) | 7 | 16 | 119% | 170ms |

`64` is the largest size that still keeps zero-window events (`zwin`) at baseline-noise levels. Beyond 64, the receiver runtime stalls become visible. Smaller chunks work too but lose useful parallelism on faster disks.

## Phases-bench results (N=2)

| Phase | A1 Δ wall | A2 Δ wall | avg | A1 utoo σ | A2 utoo σ |
|---|---|---|---|---|---|
| **p0** | −0.23s | −0.13s | **−0.18s** | 0.92 | 0.36 |
| p1 | +0.05 | +0.07 | +0.06 | 0.10 | 0.11 |
| **p3** | −0.58s | −1.58s | **−1.08s** | 1.85 | 3.15 |
| p4 | −0.05 | −0.09 | −0.07 | 0.02 | 0.05 |

For comparison, the prior `par_iter` zero-copy form had `utoo σ` of 1.70 / 1.42 (vs `par_chunks(64)` 0.36-0.92) and `p0 Δ wall` averaging +1.18s (vs −0.18s here).

A2 also marked the first run where utoo's `p0_full_cold` mean (8.04s) beat bun's (9.88s).

## Scope

- `crates/pm/src/util/extractor.rs` only.
- Integration tests in `pm-bench-pcap.sh` (dispatched separately) verify TCP + iostat metrics; no new unit tests warranted — extraction correctness is covered by existing `extract_and_write` tests.

## Risk

- The path-traversal / Tar Slip guards in `parse_tar_entries` are explicitly preserved.
- Buffer-overrun check at `parse_tar_entries` ensures `data_offset + data_len <= buf.len()` before constructing the slice.
- `par_chunks(64)` is a proper subset of rayon's primitives; falls back gracefully on any platform.

### DIFF
--- a/.github/workflows/pm-e2e-bench.yml
+++ b/.github/workflows/pm-e2e-bench.yml
@@ -154,7 +154,7 @@ jobs:
       - name: Build next branch utoo (bench baseline)
         if: >
           (github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'benchmark')) ||
-          (github.event_name == 'workflow_dispatch' && inputs.target == 'pm-bench-phases')
+          (github.event_name == 'workflow_dispatch' && (inputs.target == 'pm-bench-phases' || inputs.target == 'pm-bench-pcap'))
         run: |
           # The previous Upload step has already shipped the PR binary; we
           # can now overwrite target/.../release/utoo safely.
@@ -168,7 +168,7 @@ jobs:
       - name: Upload utoo-next binary
         if: >
           (github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'benchmark')) ||
-          (github.event_name == 'workflow_dispatch' && inputs.target == 'pm-bench-phases')
+          (github.event_name == 'workflow_dispatch' && (inputs.target == 'pm-bench-phases' || inputs.target == 'pm-bench-pcap'))
         uses: actions/upload-artifact@v4
         with:
           name: utoo-next-linux-x64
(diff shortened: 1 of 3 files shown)


elrrrrrrr:
Addressed in `1040690d`:

1. **`as usize` → `usize::try_from(...)`** (medium): applied. Even though utoo only ships 64-bit binaries (so the truncation path is unreachable in practice), `try_from` makes the failure mode of a malformed tar header explicit instead of silent. The buffer-overrun guard a few lines below catches the same class of issue, but belt-and-suspenders is fine for tar parsing.

2. **`is_none_or` → `map_or(true, ...)`** (medium): not applied. The project pins to `nightly-2026-04-02` via `rust-toolchain.toml`, which is well past `is_none_or`'s 1.82 stabilization. No MSRV concern here.

🤖

xusd320 (approved):

PR #2906, opened 2026-05-07
chore(bench): extend pcap to install phase + utoo-next baseline

elrrrrrrr:
## What

The pcap-only bench was previously a one-off that captured `p1_resolve` across `utoo` and `bun`, and assumed the project was already cloned by `pm-bench-phases.sh`. Install-phase regressions (#2902 / #2903 / #2904 / #2905 σ widening on `p0_full_cold`) live in the tarball download path — not in resolve.

This PR makes the pcap bench self-contained and covers both phases for three PMs.

## Script changes (`bench/pm-bench-pcap.sh`)

- Self-clone the project if `\$PROJECT_DIR` is missing (mirrors `pm-bench-phases.sh`)
- Add a `<pm>-install` capture per PM: lock pre-existing, cache + `node_modules` wiped, then `<pm> install`. This is the cold-tarball-download phase where the σ-widening lives.
- Add `utoo-next` as a third PM, gated on `\$UTOO_NEXT_BIN` so local runs still work without it.

## Workflow changes (`.github/workflows/pm-e2e-bench.yml`)

- `pm-bench-pcap-linux` now downloads the `utoo-next-linux-x64` artifact and exports `UTOO_NEXT_BIN`, exactly like `bench-phases-linux` does.
- The `Build next branch utoo` and `Upload utoo-next binary` steps in `build-linux` now also fire for `inputs.target == 'pm-bench-pcap'`, not only `pm-bench-phases`.

## Outputs

\`\`\`
/tmp/pm-bench-pcap/
  dns.txt
  utoo-{resolve,install}.{pcap,log}
  utoo-next-{resolve,install}.{pcap,log}   (when UTOO_NEXT_BIN set)
  bun-{resolve,install}.{pcap,log}
\`\`\`

## Why

Drives the analysis of whether the install hot-path's increased concurrency (FuturesUnordered streaming, zero-copy tar, TTY-gate) saturates outbound TCP and starves the download path — the suspect mechanism for σ widening on `p0_full_cold` across the #2903 / #2904 / #2905 split experiments.

## Smoke test

Dispatched on this chore branch first: utoo binary and utoo-next binary built off the same code (just bench scripts differ), so any utoo-vs-utoo-next install-phase delta on this run is pure environment noise — establishes the floor before we run on the experiment branches.

### DIFF
--- a/.github/workflows/pm-e2e-bench.yml
+++ b/.github/workflows/pm-e2e-bench.yml
@@ -154,7 +154,7 @@ jobs:
       - name: Build next branch utoo (bench baseline)
         if: >
           (github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'benchmark')) ||
-          (github.event_name == 'workflow_dispatch' && inputs.target == 'pm-bench-phases')
+          (github.event_name == 'workflow_dispatch' && (inputs.target == 'pm-bench-phases' || inputs.target == 'pm-bench-pcap'))
         run: |
           # The previous Upload step has already shipped the PR binary; we
           # can now overwrite target/.../release/utoo safely.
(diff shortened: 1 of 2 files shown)


elrrrrrrr:
pcap tooling rolled into #2924 stack (ci(pcap): install/preload/manifest captures). Archived.

PR #2910, opened 2026-05-08
perf(pm): TTY-gate + cooperative yield_now in install spawn loop

elrrrrrrr:
## Why

Companion experiment to #2903 (TTY-gate). After #2905 (zero-copy + par_chunks) merged, TTY-gate's phases bench *still* showed a stable +1.75s p3 regression on N=3, but pcap+iostat on the new arch (run [25535139270](https://github.com/utooland/utoo/actions/runs/25535139270)) ruled out disk and TCP path:

- `tcp.analysis.zero_window` events: **utoo 16 vs utoo-next 2** — receiver back-pressure
- `retransmits`: 16 vs 10 (baseline range)
- `gap_p99`: 1.98ms vs 2.5ms (TCP itself fast)
- `util_max` / `w_await`: identical disk profile

So the regression is **tokio runtime starvation**, not disk burst.

## Hypothesis

Before #2903, every iteration of the synchronous spawn loop in `install_packages` called `PROGRESS_BAR.inc(1)`. indicatif's internal mutex+state update gave the runtime a per-package microsecond pause — enough on a single-threaded executor to interleave socket reads on in-flight tarball downloads. Removing the call eliminated the implicit yield; the spawn loop now runs uninterrupted through ~2000 packages per layer, the kernel recv buffer fills before tokio drains it, server emits zero-window pauses.

## Change

```rust
const YIELD_EVERY: usize = 32;
// ...
for (path, package) in packages.iter() {
    iter_count += 1;
    if iter_count.is_multiple_of(YIELD_EVERY) {
        tokio::task::yield_now().await;
    }
    // ... existing logic ...
}
```

Restores the implicit yield without re-introducing the indicatif mutex cost.

## Decision matrix

| Outcome | Verdict |
|---|---|
| p3 Δ collapses to ≈0 + zwin drops to ≤5 | hypothesis confirmed; ship the yield + the TTY-gate together |
| zwin stays high, p3 still +1.5s+ | mechanism is elsewhere (spawn pace itself); consider throttling or close #2903 |
| zwin drops but p3 still bad | yield works at TCP layer but something else is the wall-time bottleneck — investigate |
| Over-corrects (p3 worse) | tune YIELD_EVERY (64 / per-iter) |

No merge intent — pure observation experiment. If signal looks good, fold the change into #2903 and graduate together.

### DIFF
--- a/.github/workflows/pm-e2e-bench.yml
+++ b/.github/workflows/pm-e2e-bench.yml
@@ -154,7 +154,7 @@ jobs:
       - name: Build next branch utoo (bench baseline)
         if: >
           (github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'benchmark')) ||
-          (github.event_name == 'workflow_dispatch' && inputs.target == 'pm-bench-phases')
+          (github.event_name == 'workflow_dispatch' && (inputs.target == 'pm-bench-phases' || inputs.target == 'pm-bench-pcap'))
         run: |
           # The previous Upload step has already shipped the PR binary; we
           # can now overwrite target/.../release/utoo safely.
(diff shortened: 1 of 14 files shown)


elrrrrrrr:
Closing — superseded by #2911. The explicit yield_now hack confirmed the runtime-starvation hypothesis but #2911 fixes it structurally by partitioning the install loop, making yield_now redundant.

PR #2915, opened 2026-05-08
perf(pm): TTY-gate progress + cooperative yield in spawn loop

elrrrrrrr:
## What

Smaller alternative to #2911. Two install hot-path changes layered on top of the new (post-#2905) zero-copy + par_chunks(64) baseline:

1. **TTY-gate progress bar** (carried from `experiment/install-tty-gate` parent commit): non-TTY callers skip indicatif's internal mutex, removing ~9k atomic ops per install.

2. **Single `consume_budget().await` at the top of `install_packages` spawn loop**: gives the tokio runtime a per-iteration window to drain socket reads on in-flight tarball downloads, replacing the implicit yield that the indicatif mutex used to provide.

```rust
for (path, package) in packages.iter() {
    tokio::task::consume_budget().await;  // ← only addition
    // ... existing logic, unchanged ...
}
```

## Why this not the bigger #2911 partition

#2911's classify + 3-phase pipeline did fix the TCP-level starvation (zwin 14 → 0), but **N=4 phases bench measured a consistent +1.05s p3 regression**. Mechanism: the partition design pushed all cheap paths (omit / cpu-incompat / `file:` link) before any spawns, then opened Phase 3 with 64 in-flight downloads simultaneously. That concentrated disk burst widened p3 σ vs the baseline's interleaved cheap+heavy schedule.

This PR keeps the **original interleaved schedule** (cheap paths inline with spawns) and only adds the runtime-yield hint. Same TCP fix, smaller change, no concentrated burst.

## Pcap mechanism evidence (carried over from partition experiments)

| Variant | utoo zwin | utoo-next zwin |
|---|---|---|
| TTY-gate alone (no yield) | 14-16 | 0 |
| TTY-gate + cooperative yield (this PR) | **0** | **0** |

`consume_budget` was added in tokio 1.41 and is the right primitive for this case: ~5ns when it doesn't yield (the common case), ~100ns when the per-task tick budget exhausts. utoo pins tokio 1.51.

## Companion / supersedes

- Supersedes #2911 (partition design) — same TCP fix at smaller surface area, without the disk-burst-induced p3 regression.
- Supersedes #2903 (TTY-gate alone) — keeps TTY-gate, fixes its known +1.75s p3 regression with one line.


elrrrrrrr:
Superseded by #2924 / #2929 stack — experiment archived.

PR #2916, opened 2026-05-08
perf(pm): bump manifests-concurrency-limit 64 → 256 for p1_resolve

elrrrrrrr:
## What

Cuts utoo's p1_resolve wall to close the gap with bun. Two-part change:

1. **Instrumentation** (\`crates/ruborist/src/util/timing.rs\`): per-fetch atomic accumulator that records (request, body, parse) split timings inside \`fetch_full_manifest\` + \`fetch_version_manifest\`. Dumped at INFO level after preload + bfs phases.

2. **Concurrency bump**: \`manifests-concurrency-limit\` default 64 → 256, matching bun's observed working point against npmjs.org.

## Evidence (local Mac, ant-design, cold)

\`\`\`
preload_wall = 22378ms
bfs_wall     = 236ms     (1% of p1)

fetch-timings: n=2730
  sum_request   = 1089s    (88% — TCP+TLS+HTTP RTT)
  sum_body      = 138s     (11% — body)
  sum_parse     = 2s       (0.16% — JSON)
\`\`\`

Effective parallel during preload was 1089s / 22.4s ≈ 49× — the 64 cap was at saturation. Per-request RTT dominates; the only lever is more in-flight requests.

Pcap from prior PRs:
- bun: ~260 TCP streams on resolve, p1 wall 2.3s
- utoo @ cap=64: ~70 streams, p1 wall 3.2s

## Expected on GHA

CI phases bench should show p1_resolve drop toward bun's range. The fetch-breakdown lines in CI logs will tell us exactly how the 64→256 bump redistributes time.

## Risk

- npmjs.org has demonstrated it accepts 260+ concurrent requests from one origin (bun proves it). 256 has headroom.
- Antgroup's npm mirror also configured to support large concurrency in past benchmarks.
- The new INFO-level breakdown lines are safe to leave in — they fire once per resolve, atomic recording is lock-free.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2918, opened 2026-05-09
experiment(pm): swap DiskManifestStore for NoopStore (measurement-only)

elrrrrrrr:
Side experiment branched off perf/p1-resolve-concurrency at 01d15130.

Question under test: how much of utoo's p1/p3 wall comes from the per-fetch disk-cache existence-check IO that `service::registry::UnifiedRegistry` issues alongside each manifest fetch.

Swaps `Context::manifest_store` from `DiskManifestStore` to `NoopStore`. NOT for landing — measurement-only branch. Compare against #2916 to read the disk-cache IO cost.


elrrrrrrr:
Measurement-only experiment — closed, signal captured in #2924 architecture notes.

PR #2920, opened 2026-05-10
perf(pm): isolate resolve hot path from install pipeline contention

elrrrrrrr:
## Summary

Targets the p0/p1 outlier tail seen on \`perf/p1-resolve-concurrency\`
(PR #2916). Mechanism: on 2-core GHA runners, mb_fetch_with_graph
main loop, graph_worker, and the install pipeline (download + clone
+ extract) all share tokio's worker thread pool. Under burst load,
graph_worker monopolizes a worker for tens of ms at a time, starving
the main loop's socket polling — eff_par_full collapses 73-77 → 40,
mb_fetch wall jumps 4-6s → 10s+, p0 tail blows out by 3-5s.

Two commits, designed to be A/B'd individually:

- **\`floor tokio worker_threads at 4\`** — bumps the per-process worker
  count from \`num_cpus\` (= 2 on GHA) to a floor of 4. Cheapest fix;
  no-op on 4+ core machines.
- **\`move graph_worker to blocking pool\`** — converts graph_worker
  from \`async fn\` (\`tokio::spawn\`) to sync \`fn\`
  (\`tokio::task::spawn_blocking\`). graph_worker is pure CPU + channel
  IO, doesn't need worker scheduling, and the blocking pool has 512
  default slots so it never contends.

## Test plan
- [ ] bench-phases-linux on 21d9c7dc (parent) → this branch tip: confirm p0 σ drops + utoo p0 mean ≤ utoo-next
- [ ] bench-phases-mac: same
- [ ] e2e green (no regression in alias / install paths)
- [ ] Optional: pcap on p0_full_cold to confirm zwin events drop on the affected runs

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2922, opened 2026-05-10
perf(pm): Plan B — inline mb_fetch + per-iter spawn_blocking (vs PR #2920 channel)

elrrrrrrr:
A/B branch from origin/next testing whether the channel rewrite in PR #2916/#2920 was overkill. Brings in PR #2920's full work but rewrites the hot loop to remove channels: graph state owned by main task, shipped into spawn_blocking per iteration. Tradeoff: per-iter dispatch overhead vs simpler code. Bench result will determine if channel arch is justified.

Generated with Claude Code.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1691,23 +1691,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "ctor"
-version = "0.10.1"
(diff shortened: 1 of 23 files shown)


elrrrrrrr:
Plan B vs #2920 — channel-based approach won, see #2924.

PR #2925, opened 2026-05-10
perf(pm): candidate D — skip preload, BFS-only minimal change

elrrrrrrr:
Minimal change on origin/next: skip the standalone preload phase, let BFS resolve on demand via OnceMap+UnifiedRegistry. Adds worker_threads ≥ 4. No service-layer rewrite. A/B vs PR #2920 to measure if preload's wave-shape parallelism is necessary or if BFS-on-demand catches up via OnceMap single-flight.

### DIFF
--- a/crates/pm/src/main.rs
+++ b/crates/pm/src/main.rs
@@ -344,9 +344,15 @@ enum Commands {
 fn main() {
     crate::util::sysconf::init();
 
+    // Floor at 4 worker threads even on 1-2 core CI runners. Without
+    // this the install pipeline (resolve + download + clone) competes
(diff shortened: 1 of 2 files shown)


elrrrrrrr:
Falsified: BFS is fully sequential per-edge (`for edge in unresolved { process_dependency().await }` in builder.rs:917). Without preload's parallel prefetch warming the cache, every dep blocks on its own network roundtrip → p0 87s, p1 74s vs origin/next preload+BFS p0 9.7s, p1 3.3s. The 'parallel prefetch' comment in run_bfs_phase docs refers to preload itself, not a separate per-level prefetch. Minimal change to remove preload requires also making BFS concurrent — out of scope for this exp.

PR #2924, opened 2026-05-10
perf(pm): channel install pipeline — architectural decision log + wasm fix

elrrrrrrr:
Stacks on PR #2920 (channel + install integration). Goal: ship the channel-architecture install path as the single uniform code path, capture the architectural decisions with theoretical justification + bench evidence, and remove the legacy preload+BFS code (with a wasm32 carve-out).

## Architectural decision log

For each architectural commitment, what we picked, why it works, and how we'd falsify it.

### 1. Channel-based fetch + graph separation (vs inline graph mutation)

**Picked**: `mb_fetch_with_graph` ships fetched events over `tokio::mpsc` to a separate `graph_worker` task; main loop never blocks on graph mutation between fetches.

**Theory**: graph mutation is pure CPU (~50us per dep × 4647 = ~230ms total). If inline, main loop blocks on each per-fetch graph step → next \`futs.next().await\` is delayed → effective parallelism collapses. Decoupling keeps main loop continuously polling sockets.

**Falsification (Plan B in PR #2922 sweep)**: A/B with inline graph mutation wrapped per-iteration in \`tokio::task::spawn_blocking\` to keep main loop free. Per-iteration spawn_blocking overhead (~tens of µs × 4647 ≈ 280ms) ate the channel arch's p1 win → Plan B p1 = 3.21s vs PR #2920 channel p1 = 2.78s, a 0.43s gap. Channel arch wins.

### 2. \`graph_worker\` on tokio blocking pool (vs worker scheduler)

**Picked**: \`tokio::task::spawn_blocking(move || graph_worker(...))\` not \`tokio::spawn(...)\`.

**Theory**: graph_worker is pure CPU + blocking channel I/O, no async awaits. On a multi-thread runtime the worker scheduler has only \`num_cpus\` threads (= 2 on GHA ubuntu-latest). graph_worker can monopolize a worker thread for tens of ms during burst processing, starving main loop's socket polling. Blocking pool defaults to 512 slots → reserving one for graph_worker has zero scheduling effect on download/clone tasks (which spawn_blocking via fs ops too).

**Falsification (PR #2920 commit \`279c51e8\`)**: A/B with \`tokio::spawn(graph_worker)\` showed eff_par_full collapse 73-77 → 40 on outlier runs and p0/p1 tail blowout +3-5s. Switching to spawn_blocking made p0 σ collapse from 2.98s → 0.16s (19×) on linux, 3.32s → 0.25s (13×) on mac.

[...]

PR #2926, opened 2026-05-11
perf(pm): MANIFESTS_CONCURRENCY 96→128 + TCP/TLS warmup

elrrrrrrr:
Two p1_resolve micro-opts on top of PR #2924. Target: linux p1 below 2.5s (current best ~2.42s, mean 2.80s). See commit message for theory + falsification conditions.


elrrrrrrr:
Superseded by #2929 (perf/p1-sibling-two-phase) — same p1 attack line.

PR #2929, opened 2026-05-11
perf(pm): two-phase sibling parse + 4× channel buffer (p1 attack)

elrrrrrrr:
Two micro-opts on PR #2924 stack targeting the 0.5s framework overhead vs manifest-bench. See commit message.


elrrrrrrr:

PR #2930, opened 2026-05-11
perf(pm): floor tokio worker_threads + rayon pool for small CI hosts

elrrrrrrr:
## Summary

Two small defaults tweaks for the resolve hot path on small CI hosts (GHA ubuntu-latest reports 2 cores):

- **`tokio worker_threads`**: floor at 4 so a CPU-bound task can't monopolize one of two workers and stall socket polling on the others (TCP zwin events from the remaining worker stretching tail wall).
- **`rayon` pool**: floor at 8 so the per-fetch manifest parse + extract dispatches don't queue serially behind earlier ones; oversubscribed but still bounded by host CPU.

Both floors no-op on bigger hosts (`num_cpus >= 4` already → tokio floor inert; `num_cpus >= 8` → rayon floor inert).

## Why ship these together

Both are "give the resolve/install pipeline enough thread headroom on a 2-core host" — same problem class, different runtimes. Each in isolation is small noise; together they consistently beat baseline on the GHA bench grid without hurting bigger hosts.

The original draft also bumped `manifests-concurrency-limit` from 64 → 96. Held back to a separate PR — that one wants its own bench window, and the right design probably includes an adaptive back-off informed by real throttling traces rather than a one-shot bump.

## Test plan

- [x] `cargo fmt --check` (clean on poolab)
- [x] `cargo clippy -p utoo-pm --all-targets -- -D warnings --no-deps` (clean)
- [x] `cargo build -p utoo-pm --profile release-local` (43s on poolab)
- [x] `cargo test -p utoo-pm --release` (261 passed, 3 ignored, 0 failed)
- [ ] pm-bench-phases CI: confirm no p0/p1/p3/p4 regression on next baseline (label triggered)
- [ ] e2e: \`./e2e/utoo-pm.sh\`

### DIFF
--- a/crates/pm/src/main.rs
+++ b/crates/pm/src/main.rs
@@ -344,9 +344,15 @@ enum Commands {
 fn main() {
     crate::util::sysconf::init();
 
+    // Floor worker_threads at 4 on small CI runners (GHA ubuntu-latest
+    // reports 2). Install multiplexes resolve fetch + download + clone
+    // tasks; with 2 workers a CPU-bound segment monopolizes a thread
+    // and stalls socket polling on the others, producing TCP zwin
+    // events that stretch the tail wall.
     let worker_threads = std::thread::available_parallelism()
         .map(|n| n.get())
-        .unwrap_or(4);
+        .unwrap_or(4)
+        .max(4);
 
     let result = tokio::runtime::Builder::new_multi_thread()
         .enable_all()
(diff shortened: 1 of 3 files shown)


elrrrrrrr:
Closing — bench data on GHA shows variance blow-up (σ 7-9× vs baseline) + small p1/p0 regression on 2-core. Hypothesis was that mb_resolve's graph_worker contention justified the thread floors, but origin/next's BFS resolver doesn't have that contention pattern, so the oversubscription is pure overhead. Will revisit when mb_resolve lands.

Poolab cross-check was no-signal (192-core box → both floors no-op), confirming the change only activates on small CI hosts and there's no win there.

PR #2931, opened 2026-05-11
perf(pm): fold preload + BFS into one streaming resolve pass

elrrrrrrr:
## Summary

Replace ruborist's two-phase resolve (parallel preload then sequential BFS over warm cache) with a **single streaming pass**: `FuturesUnordered` drives registry fetches at `config.concurrency`, graph mutation runs inline on the main loop as each fetch lands.

Eliminates the ~500-1000ms warm-cache walk the BFS phase did on large workloads (ant-design ≈ 4.6k packages).

## Before — two phases, sequential

```
   ┌──────────────────────────────────────┐
   │ run_preload_phase                    │
   │                                      │
   │   pending ──▶ FuturesUnordered       │   network parallel
   │                  │                   │
   │                  ▼                   │
   │            populate MemoryCache      │
   │                  │                   │
   │                  ▼                   │
   │      emit PackageResolved ───────────┼──▶ tgz pipeline (early)
   └──────────────────┬───────────────────┘
                      │
                      ▼  wait for all fetches
   ┌──────────────────────────────────────┐
   │ run_bfs_phase                        │
   │                                      │
   │   for level in BFS:                  │   sequential walk
   │     for node in level:               │
   │       for edge in unresolved:        │
   │         process_dependency().await   │   cache hit, just mutate
   │                                      │   ~500-1000ms wall, no I/O
   │         emit Resolved + PackagePlaced│
   └──────────────────────────────────────┘
```

The BFS phase is fully sequential per-edge. On a warm cache, every `process_dependency().await` is just a graph mutation (no network, no CPU). The await cost dominates the wall — pure overhead.

## After — one streaming pass

```
   ┌──────────────────────────────────────────────┐
   │ run_bfs_phase (streaming fold)               │
   │                                              │
   │   pending ◀──┐                               │
   │      │       │ push new transitive           │
   │      ▼       │ edges                         │
   │   dispatch   │                               │
   │      │       │                               │
   │      ▼       │                               │
   │   FuturesUnordered (96-cap)                  │   network parallel
   │      │                                       │
   │      ▼  fetch lands                          │
   │   process_dependency(pre_resolved)           │   inline graph mutation
   │      │                                       │
   │      ├──▶ emit PackageResolved ──────────────┼──▶ tgz pipeline
   │      ├──▶ emit Resolved + PackagePlaced      │
   │      └──▶ collect new edges ─────────────────┘
   │                                              │
   │   (Reuse fast-path + non-registry edges      │
   │    short-circuit to inline handling          │
   │    without going through the futures pool)   │
   └──────────────────────────────────────────────┘
```

Same fetch-loop shape as the old preload, plus inline graph mutation when each fetch lands. The transitive walk that BFS used to do is now driven by the same loop — every Created node's new edges go straight back into `pending`.

## What stays the same

- **Same `BuildEvent` stream**: `PreloadStart/Queued/Fetching/Progress`, `PackageResolved`, `Resolving/Resolved/Reused/Skipped`, `PackagePlaced` — install pipeline (tarball downloads triggered by `PackageResolved`) and progress bar work unchanged.
- **Same correctness**: non-registry edges (workspace/git/http/file) and Reuse hits short-circuit to inline handling, identical to before.
- **Same concurrency knob**: `manifests-concurrency-limit` default unchanged.
- **No new modules / functions / public API**: refactor is in-place.

## What gets deleted

- `run_preload_phase` (its work is now folded into the main streaming loop)
- `gather_preload_deps` (only caller was `run_preload_phase`)
- `BuildDepsConfig::skip_preload` field + `with_skip_preload` method (no preload phase to skip)
- 2 unit tests that were probing graph state via `gather_preload_deps`

## Net diff

```
crates/ruborist/src/resolver/builder.rs   +319 -308   (+11 net)
crates/ruborist/src/service/api.rs        +0   -9     (-9 net)
```

## Test plan

- [x] `cargo fmt --check` clean (poolab)
- [x] `cargo clippy --all-targets -- -D warnings --no-deps` clean (poolab)
- [x] `cargo build --profile release-local` (14s on poolab)
- [x] `cargo test -p utoo-pm --release`: 261 passed, 3 ignored, 0 failed
- [x] `cargo test -p utoo-ruborist --lib`: 163 passed, 0 failed
- [ ] pm-bench-phases on **poolab**: p0/p1/p3/p4 vs origin/next baseline (in flight)
- [ ] pm-bench-phases on **GHA**: p0/p1/p3/p4 vs origin/next baseline (CI label triggered)
- [ ] e2e: `./e2e/utoo-pm.sh`

### DIFF
--- a/crates/ruborist/src/model/graph.rs
+++ b/crates/ruborist/src/model/graph.rs
@@ -12,7 +12,7 @@ use super::manifest::{CoreVersionManifest, NodeManifest};
 use super::node::{EdgeType, NodeType};
 use super::override_rule::{OverrideRule, Overrides};
 use super::package_json::PackageJson;
-use crate::resolver::semver::matches;
+use crate::resolver::semver::{matches, normalize_spec};
 
 /// Package node in the dependency graph.
 #[derive(Debug, Clone)]
@@ -447,6 +447,18 @@ impl DependencyGraph {
     }
 
     /// Recursively search for compatible node in parent chain.
+    ///
+    /// Identity check mirrors npm's `Node.matches`: an existing slot can
+    /// satisfy the requesting edge only when both the slot name and the
+    /// underlying package identity match. Differences:
+    /// - Regular range edges only constrain version: any package in the
+    ///   slot whose version satisfies the range is a Reuse (matches
+    ///   `dep-valid.js`'s 'range' case — `semver.satisfies(child.version)`
+    ///   with no packageName check).
+    /// - `npm:` alias edges additionally require the slot's manifest
+    ///   `packageName` to equal the alias target (`subSpec` in npm parlance).
+    ///   When the slot is occupied by a different underlying package the
+    ///   alias is explicit and gets the slot via `Replace`.
     fn find_in_parent_chain(
         &self,
         current: NodeIndex,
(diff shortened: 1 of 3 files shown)


elrrrrrrr:
Closing — streaming fold experiment failed exit criteria for the perf-validation loop.

## Bench data (GHA Linux, same-run apples-to-apples)

| Phase | PR-D | utoo-next | Δ |
|---|---|---|---|
| p0_full_cold | 17.36 ± 4.07s | 13.20 ± 0.89s | +31% ❌ |
| p1_resolve | 3.19 ± 0.19s | 2.71 ± 0.10s | +18% ❌ |
| p3_cold_install | 10.82 ± 4.23s | 8.63 ± 1.51s | +25% ❌ |
| p4_warm_link | 4.78 ± 0.55s | 3.61 ± 0.36s | +32% ❌ |

σ inflation 4-5× of utoo-next is a structural signal — not noise. Even if absolute numbers vary with GHA network conditions, the relative regression is consistent.

## Why streaming fold is structurally slower here

`find_compatible_node` is called twice per registry edge:
1. Streaming loop pre-check (load-bearing for workspace dep resolution + Reuse fast path)
2. Inside `process_dependency` after fetch lands

Removing the pre-check broke workspace dep resolution (verified e2e — `workspaces-scoped-pkg` failed with 404 because `@ruyadorno/scoped-b` is a workspace package, not a registry package). The pre-check isn't redundant, it's load-bearing.

The two-pass `find_compatible_node` per edge is ~5000 × 2 parent-chain walks for ant-design. Old preload + BFS only walks once per edge (inside `process_dependency`) and benefits from a cleaner separation: preload's fetch loop has no graph mutation overhead.

## What's salvageable from this experiment

The npm-aware `find_compatible_node` fix (the `Replace` variant + `child.manifest.name()` underlying-identity check) is an architectural correctness improvement that's orthogonal to streaming fold — npm's `Node.matches` checks `packageName`, ours used to match by slot name + version only. Currently masked by BFS level ordering in origin/next; could be revisited as a defensive fix in a smaller PR.

## E2e status

- macOS aarch64 / macOS x86_64 / Windows: PASS (alias case 11e fix verified)
- Ubuntu Linux: workspace dep test FAIL on intermediate commit, PASS on full PR-D

## Follow-up

Per the autonomous loop's fallback path, this PR is closed. The streaming fold direction is set aside; if perf work resumes on the resolve hot path, the more promising direction is the channel-based architecture from #2929 with a focus on bounding code size.

PR #2933, opened 2026-05-12
perf(pm): channel-based main loop with spawned preload, BFS by level

elrrrrrrr:
## Summary

Replace the sequential `run_preload_phase().await + run_bfs_phase().await` two-phase model with a concurrent design where preload runs as a spawned task and the main loop owns graph + cache + BFS by level.

## Architecture

```
[spawned preload — autonomously running]      [main loop — sole graph + cache writer]
─────────────────────────────────────────      ────────────────────────────────────────
own pending: VecDeque<(name, spec)>            for level:
own FuturesUnordered (cap satured)               for edge in level:
walks transitives by manifest content:             cache hit (preload 已 fetch) → process inline
  for each fetched manifest:                       cache miss → defer
    extract deps → push to pending                 (later) drain mpsc → write cache
parallel fetch via fetch_full_manifest                               → process deferred edges
on result: send via mpsc ──────────────►       swap current/next level
```

**Properties**:
- **Single-writer cache**: main owns local `HashMap<String, Arc<FullManifest>>`. Eliminates DashMap shard contention from concurrent preload fetch tasks.
- **Strict BFS level barrier**: level N edges fully resolved before level N+1 starts. Preserves npm-aligned `npm:` alias slot occupancy without needing a `Replace` graph fix.
- **Cache keyed by underlying name**: `npm:raw-body@2.1.3` and a real `ms` package fetch into separate cache slots; race between alias and same-named real package can't poison either slot.
- **Preload bandwidth filled regardless of BFS progress**: preload runs autonomously at `config.concurrency` cap; doesn't wait for BFS.
- **WASM falls back** to existing `run_preload_phase + run_bfs_phase`.

## Trade-offs

- **Trait change**: `RegistryClient` future-returning methods now require `+ Send` for `tokio::spawn`. Implementations need to be `Send`-compatible (`UnifiedRegistry` already is via `Arc<...>` internals).
- **R: Clone bound** added to `build_deps`/`build_deps_with_config` etc — registry is Arc-cloned into the spawned preload task.

## Future (not in this PR)

- **Priority queue** inside dispatcher: BFS-needed manifests highest priority, preload transitives next, tgz prefetch lowest. Today preload's pending is FIFO; priority becomes important when we add tgz prefetch.
- **optionalDeps allowed-fail**: currently optional fetch failures are skipped; tighter handling for partial-failure scenarios.

## Net diff

```
crates/ruborist/src/resolver/builder.rs   +574 -16
crates/ruborist/src/resolver/preload.rs   +5   -2     (helpers exposed pub(crate))
crates/ruborist/src/traits/registry.rs    +3   -1     (+Send on fetch_full_manifest future)
```

## Test plan

- [x] `cargo build -p utoo-pm --profile release-local` (poolab, 14s)
- [x] `cargo test -p utoo-ruborist --lib` (163 passed)
- [x] `cargo test -p utoo-pm` (253 passed, 3 ignored)
- [x] `cargo clippy --all-targets -- -D warnings --no-deps` clean
- [x] Smoke test: e2e case 11e (`npm:` alias) on poolab Linux — `top-level ms = raw-body` ✓, no nested `ms` ✓
- [ ] pm-bench-phases CI: target ≥ #2929 perf level (p1=2.63s, p0=7.94s, p3=7.32s) (label triggered)
- [ ] full e2e on poolab + GHA


elrrrrrrr:

PR #2935, opened 2026-05-12
perf(pm): level-dispatch registry resolution

elrrrrrrr:
## Summary
- add a native level-dispatch resolver path for registry dependencies
- centralize graph mutation and registry cache writes in the main level loop
- fetch missing full manifests through an uncached registry hook to bypass OnceMap contention
- keep the existing two-phase path for wasm

## Verification
- cargo check -p utoo-ruborist
- cargo clippy -p utoo-ruborist --all-targets -- -D warnings --no-deps
- cargo test -p utoo-ruborist

Draft PR to trigger CI/benchmark for perf comparison.


elrrrrrrr:

PR #2941, opened 2026-05-13
fix(pm): print registry auto-select hint on stderr

elrrrrrrr:
## Summary

The registry auto-select hint —

    Registry: <url> (<latency>ms)
    Tip: ut config set registry <url> --global

was printed with `println!`, putting it on **stdout**. When `ut`/`utx` output is piped or captured, the banner corrupts the data stream. The motivating regression:

    pbpaste | utx prettier --parser babel | pbcopy

prepended the two lines to prettier's stdin, breaking the round-trip. Same failure shape for `out=$(ut …)`, `ut … > file`, and `ut … --json | jq`.

This moves the two lines to `eprintln!` (**stderr**), the convention every long-lived CLI follows — cargo/git/npm/curl all emit status chatter on stderr and reserve stdout for data. The hint stays fully visible in interactive terminals and CI logs (both fds are captured); it just stops polluting fd 1.

## Intentionally unchanged

`service/config.rs`'s `println!` stays — that line is the answer to `ut config get registry`, the value a script captures via `$(...)`, so it belongs on stdout.

## Test plan

- [x] `cargo fmt` clean (same-width macro swap, no layout change).
- [ ] Interactive `ut install` with no configured registry still shows the colored `Registry:`/`Tip:` pair on the terminal.
- [ ] `pbpaste | utx prettier --parser babel | pbcopy` round-trip returns only the formatted code.

## Follow-ups (out of scope)

- Persist the auto-picked registry to global config so the banner fires once per machine instead of every invocation.
- Migrate the lingering `atty::is(...)` call to std `IsTerminal` (drop the `atty` dep).


yuzheng14 (approved):
LGGGGGGGGGGGGGTM!!!

PR #2942, opened 2026-05-14
ci(pm): reduce phase bench default runs

elrrrrrrr:
## Summary

- Lower phase-isolated PM benchmark default runs from 3 to 2.
- Apply the same default to Linux and macOS phase benchmark jobs.
- Keep explicit `bench_runs` workflow_dispatch input behavior unchanged.

## Why

This is the first split from the PM resolver/install scheduling experiment. It keeps the benchmark-label feedback loop fast enough for iterative performance work while preserving a reproducible GHA signal.

## Verification

- `git diff --check`

This PR only changes GitHub Actions benchmark defaults.

### DIFF
--- a/.github/workflows/pm-e2e-bench.yml
+++ b/.github/workflows/pm-e2e-bench.yml
@@ -73,7 +73,7 @@ on:
       bench_runs:
         description: '[phases] Runs per phase'
         required: false
-        default: '3'
+        default: '2'
         type: string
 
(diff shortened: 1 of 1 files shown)


elrrrrrrr:
Skipping this split: the run-count default change is too small to justify a standalone review PR. We can fold it into the benchmark/tuning PR later if needed.

PR #2948, opened 2026-05-14
perf(pm): add resolver demand bfs scheduler

elrrrrrrr:
## Summary

This PR moves PM resolution/install toward the mainloop-owned scheduling model we have been validating:

- Resolver BFS loop owns traversal, per-run manifest cache, waiters, and manifest job de-dupe.
- Spawned provider jobs only do concrete work: fetch manifest bytes, parse JSON, extract per-version data, and optional persistence.
- Install now has an `InstallScheduler` mainloop that owns tarball download, cache-path resolution, clone queues, clone waiters, parent blocking, and inflight de-dupe.
- The old `service/pipeline` layer was deleted. Ruborist build events now flow directly into `InstallEventReceiver`, which lives beside the scheduler.
- PM util download/clone APIs are now primitive operations: worker tasks download/extract or clone from a resolved cache path, but do not own global inflight state.
- Old ruborist `MemoryCache` / `PackageCache` / `CacheStats` global DashMap state was removed. The resolver mainloop owns per-run memory state; `service/cache.rs` only keeps serializable project-cache data.
- Exact-version manifests parse from `Vec<u8>` so the hot path avoids an extra `Bytes -> Vec` copy before `simd_json` mutates the buffer.
- Speculative full-manifest extraction is kept because it reduces p1 involuntary ctx in GHA, even though it is not the whole p1 wall-time story.

The deps `NoopStore` experiment (`747eea5f`) was useful as an AB probe but has been reverted in `a5461132`, because it changes persistence semantics and can interfere with p0/p3. Current code keeps disk-backed `ManifestStore`; the next optimization should target where persistence runs, not whether it exists.

## Current Architecture

- Mainloop: BFS expansion, manifest inflight/cache ownership, install download/clone inflight ownership, parent ordering, and waiter fanout.
- Tokio tasks: HTTP requests, tarball extract/write, clone/hardlink work, cache-path probes, and host persistence tasks.
- Rayon/global CPU pool: manifest JSON parse and per-version extraction.
- PM service layer: command orchestration only; install scheduling state lives in `InstallScheduler`.
- Util layer: atomic filesystem/network operations only, with no install hot-path `OnceMap` ownership.
- Persistent manifest/project cache: still present; write placement remains a follow-up optimization.

`util/cloner.rs`, `util/downloader.rs`, and `ruborist::service::registry` are now primitive operations/providers. They do not own resolver/install inflight state. PM `util/registry.rs` remains only for `ut ping` and `registry=auto` selection.

## Public Validation

[...]

PR #2957, opened 2026-05-15
perf(pm): experiment bounded install extract backlog

elrrrrrrr:
Experiment C for #2948.\n\nChange:\n- Limit download lead over extract/write to extract_limit * 4 instead of the full download_limit.\n\nHypothesis:\n- p0/p3 may reduce RSS, minor faults, and context switching by preventing too many downloaded tarball byte buffers from queueing ahead of extract/write.\n\nValidation target:\n- GHA linux npmjs benchmark via labels.\n- Private poollab npmmirror AB stays out of public PR comments.


elrrrrrrr:

PR #2961, opened 2026-05-18
chore(pm): add install scheduler stats

elrrrrrrr:
## Summary

Diagnostic PR on top of #2948 for the p0/p3 install loop.

Adds opt-in install scheduler stats behind `UTOO_INSTALL_SCHEDULER_STATS=1`. Default behavior does not enable timing, so normal installs and public benches should stay behaviorally unchanged.

When enabled, the scheduler logs:

- seeded-cache probe count/time and hit/miss split
- download count/time and cache/network split
- downloaded byte volume
- extract count/time
- clone count/time

## Why

p1 now has a strong wall/ctx result after the queued ManifestStore writer. p0/p3 still need phase-level attribution before deeper changes. This gives us the same style of data for install scheduler work without adding global state to worker tasks.

## Local validation

- `cargo fmt`
- `cargo test -p utoo-pm service::install_scheduler::tests`

### DIFF
--- a/crates/pm/src/service/install_scheduler.rs
+++ b/crates/pm/src/service/install_scheduler.rs
@@ -1,5 +1,6 @@
 use std::collections::{HashMap, HashSet, VecDeque};
 use std::path::PathBuf;
+use std::time::{Duration, Instant};
 
 use anyhow::{Context, Result, anyhow};
 use bytes::Bytes;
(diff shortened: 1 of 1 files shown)


elrrrrrrr:
Local stats smoke test passed with `UTOO_INSTALL_SCHEDULER_STATS=1` on a temporary `is-number@7.0.0` project. Output included scheduler phase data, for example: `seeded_cache_count=1`, `download_count=1`, `extract_count=1`, `clone_count=1`, and per-stage milliseconds. This confirms the opt-in diagnostics are usable without changing the default bench path.

elrrrrrrr:
### GHA default-path read

Run `26010586285` completed for linux/npmjs with stats disabled by default.

| phase | utoo | ctx | read |
| --- | ---: | ---: | --- |
| p0 full cold | `8.88s ±1.96` | `137.5K/90.1K` | noisy; not a perf candidate |
| p1 resolve | `2.43s ±0.10` | `15.8K/19.0K` | low-ctx shape preserved |
| p3 cold install | `6.98s ±2.18` | `125.7K/77.3K` | noisy; diagnostic only |
| p4 warm link | `2.28s ±0.04` | `53.7K/25.2K` | stable |

Read: keep this as an opt-in diagnostic PR only. It is not intended to be folded as a performance optimization unless we decide the stats hook is worth keeping for future p0/p3 attribution.

elrrrrrrr:

PR #2960, opened 2026-05-18
perf(pm): skip prefetched install cache probes

elrrrrrrr:
## Summary

AB experiment on top of #2948.

When the resolver has already emitted `PackageResolved`, the install scheduler has a registry download state for the same package. `PackagePlaced` clone requests can attach to that state directly instead of spawning a separate seeded-cache probe task first.

This keeps the fallback path unchanged for packages without registry prefetch state: file/http/git seeded cache lookups still use `resolve_seeded_cache_path`.

## Expected signal

- p0/p3 may drop scheduler tasks and ctx by avoiding redundant cache probe work on registry packages.
- p4 should be mostly neutral because it starts from warm cache and still needs clone/link work.

## Local validation

- `cargo fmt`
- `cargo clippy -p utoo-pm --all-targets -- -D warnings --no-deps`
- `cargo test -p utoo-pm service::install_scheduler::tests`
- `cargo test -p utoo-pm` (`261 passed`, `3 ignored`)

PR #2962, opened 2026-05-18
perf(pm): prefetch fresh lock downloads

elrrrrrrr:
## Summary

AB experiment for p3 fresh-lock installs.

When `utoo install` can reuse a fresh `package-lock.json`, the resolver stage is skipped, so the install scheduler does not receive `PackageResolved` events and cannot start registry downloads until each depth-level clone asks for them.

This PR prefetches registry downloads from the fresh lockfile before clone work starts:

- only considers lock entries whose incoming dependency spec is provably registry-style (`semver`, dist-tag, npm alias parsed as registry by ruborist)
- skips `file:`, `git`, generic `http(s)` tarball specs, links, unknown package names, and platform-incompatible packages
- keeps clone placement and parent ordering unchanged; this only moves registry download/cache/extract work earlier

## Why this is safe

Generic HTTP tarballs must continue to use the `_http_<url-hash>` seeded cache slot. The prefetch classifier checks the parent dependency spec, not only the `resolved` URL, so `"foo": "https://example.com/foo.tgz"` is not routed into `<name>/<version>` registry cache.

## Expected impact

Target phase: `p3_cold_install`.

This should overlap more registry download/extract work with the early install pipeline for fresh-lock installs. It should not materially affect `p1_resolve`; `p0_full_cold` already gets resolver-driven prefetch from `PackageResolved` events.

## Validation

- `cargo fmt`
- `cargo test -p utoo-pm` (`262 passed`, `3 ignored`)

## Bench plan

Draft AB PR. Trigger linux/npmjs phase bench via labels, compare against #2948 folded baseline and #2960/#2961 records.

### DIFF
--- a/crates/pm/src/service/install.rs
+++ b/crates/pm/src/service/install.rs
@@ -18,13 +18,14 @@ use crate::model::package::PackageInfo;
 use crate::service::rebuild::RebuildService;
 use crate::util::cli_enum::{OmitType, PackageAction, SaveType};
 use crate::util::cloner::clone_count;
-use crate::util::downloader::download_stats;
+use crate::util::downloader::{download_stats, is_registry_tarball_url};
 use crate::util::json::load_package_lock_json_from_path;
 use crate::util::linker::link;
 use crate::util::logger::{
     PROGRESS_BAR, finish_progress_bar, log_progress, print_install_counts, start_progress_bar,
 };
 use utoo_ruborist::compat::{is_cpu_compatible, is_os_compatible};
+use utoo_ruborist::spec::SpecStr;
 
 use super::binary::update_package_binary;
 use super::clean::clean_deps;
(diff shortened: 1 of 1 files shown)


elrrrrrrr:
### GHA run 1 read

Run `26012018790` completed for linux/npmjs.

| phase | utoo | ctx | read |
| --- | ---: | ---: | --- |
| p0 full cold | `7.96s ±0.21` | `116.8K/92.9K` | good wall, but this patch should not primarily affect p0 because cold install already gets resolver-driven prefetch |
| p1 resolve | `2.44s ±0.02` | `16.3K/20.0K` | low-ctx shape preserved; unrelated to this patch |
| p3 cold install | `6.17s ±0.05` | `106.6K/83.6K` | target phase, but no clear win vs utoo-next `6.15s` / utoo-npm `5.80s`; iCtx is higher than both baselines |
| p4 warm link | `2.27s ±0.08` | `52.7K/24.5K` | wall neutral vs utoo-next, ctx still higher |

Conclusion: do **not** fold yet. The classifier is safe, but the data says fresh-lock download start timing is not the main p3 bottleneck. Next direction should move down the install hot path: seeded-cache probe / registry cache lookup / clone-hardlink scheduling and queue shape.

elrrrrrrr:

PR #2964, opened 2026-05-18
perf(pm): skip fresh lock registry cache probes

elrrrrrrr:
## Summary

AB experiment for p3 fresh-lock install hot path.

This builds on the fresh-lock registry classifier from #2962. #2962 showed that merely starting downloads earlier does not materially improve p3. This PR goes one step lower in the hot path: when a lockfile package is proven to come from a registry spec, its clone request can directly join the registry cache/download state instead of spawning a seeded `_http_<url-hash>` cache probe first.

Changes:

- add `CloneCacheMode::Registry` to the install scheduler
- fresh-lock registry packages use `ensure_registry_clone`, which routes directly to `ensure_download`
- normal `ensure_clone`, resolver clone prefetch, git/file/generic HTTP tarballs still use the seeded-cache path
- keep parent/depth clone ordering unchanged

## Safety boundary

The registry fast path is only used for lock entries whose incoming parent dependency spec parses as a ruborist registry spec. Generic `https://...tgz`, `file:`, git, links, unknown package names, and platform-incompatible packages are excluded and continue through seeded-cache lookup.

## Validation

- `cargo fmt`
- `cargo test -p utoo-pm service::install_scheduler::tests`
- `cargo test -p utoo-pm fresh_lock`
- `cargo test -p utoo-pm` (`263 passed`, `3 ignored`)

## Bench plan

Compare against #2962 and #2948 on linux/npmjs. Target phase is `p3_cold_install`; expected signal is lower ctx and possibly lower wall. p4 should stay neutral.

### DIFF
--- a/crates/pm/src/service/install.rs
+++ b/crates/pm/src/service/install.rs
@@ -18,13 +18,14 @@ use crate::model::package::PackageInfo;
 use crate::service::rebuild::RebuildService;
 use crate::util::cli_enum::{OmitType, PackageAction, SaveType};
 use crate::util::cloner::clone_count;
-use crate::util::downloader::download_stats;
+use crate::util::downloader::{download_stats, is_registry_tarball_url};
 use crate::util::json::load_package_lock_json_from_path;
 use crate::util::linker::link;
 use crate::util::logger::{
     PROGRESS_BAR, finish_progress_bar, log_progress, print_install_counts, start_progress_bar,
 };
 use utoo_ruborist::compat::{is_cpu_compatible, is_os_compatible};
+use utoo_ruborist::spec::SpecStr;
 
 use super::binary::update_package_binary;
 use super::clean::clean_deps;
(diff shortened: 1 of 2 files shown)


elrrrrrrr:
### GHA run 1 read

Run `26012701499` completed for linux/npmjs.

| phase | utoo | ctx | read |
| --- | ---: | ---: | --- |
| p0 full cold | `7.63s ±0.07` | `126.3K/94.8K` | good wall in this run, ahead of utoo-next/npm/bun, but this patch targets fresh-lock clone path rather than p0 directly |
| p1 resolve | `2.33s ±0.05` | `14.9K/20.2K` | low-ctx shape preserved; unrelated to this patch |
| p3 cold install | `8.01s ±3.70` | `144.8K/68.9K` | target phase did not improve; worse/noisier than #2962 `6.17s ±0.05` and not better than #2948 folded `7.62s ±0.85` |
| p4 warm link | `2.23s ±0.07` | `51.3K/24.0K` | roughly neutral vs recent runs |

Conclusion: do **not** fold this as-is. Skipping the seeded `_http_<url-hash>` probe for fresh-lock registry clones does not show a p3 win on GHA; the bottleneck is likely further down the install path, around clone/link queue shape, hardlink/symlink filesystem work, or residual scheduler coupling rather than registry cache lookup alone.

elrrrrrrr:

PR #2965, opened 2026-05-18
perf(pm): run install clone workers on rayon

elrrrrrrr:
### What

AB experiment for p3/p4 install/link ctx: move scheduler-owned clone workers from the nested `tokio::spawn -> spawn_blocking` path to a single rayon worker that reports completion back to the install scheduler main loop.

### Why

Current p4 is mostly warm-cache clone/link work. Each package clone currently enters a tokio task and then hops again into the blocking pool for sync hardlink/copy. This tests whether removing that extra async/blocking-pool hop reduces context switches without changing scheduler ownership, clone ordering, cache lookup, or package semantics.

### Notes

- Scheduler state remains in the main loop.
- Clone workers still only perform atomic filesystem work and send an `OpDone::Clone` result back.
- Existing async `clone_package` remains for non-scheduler call sites.
- This is an AB PR; fold only if p3/p4 GHA data improves.

### Validation

- `cargo fmt`
- `cargo test -p utoo-pm`
- `cargo clippy -p utoo-pm --all-targets -- -D warnings --no-deps`

### Bench plan

Trigger linux/npmjs phase bench and compare mainly:

| phase | baseline #2948 | target |
| --- | ---: | --- |
| p3 cold install | `7.62s`, ctx `142.2K/68.3K` | wall neutral/better, ctx lower |
| p4 warm link | noisy `4.83s`, ctx `54.6K/23.5K`; recent ABs around `2.2s`, ctx `~52K/~24K` | ctx lower without wall regression |

PR #2966, opened 2026-05-18
perf(pm): handle warm registry cache hits in scheduler

elrrrrrrr:
### What

AB experiment for p4 warm-link ctx: let the install scheduler handle registry cache hits synchronously in the main loop before enqueueing download work.

### Why

In p4 the lockfile and registry cache are already warm, but every registry package still enters a worker task just to check `<cache>/<name>/<version>/_resolved`. This keeps the scheduler state centralized while removing one per-package tokio worker from the all-cache-hit path.

### Notes

- Cache miss behavior is unchanged: misses still enqueue download work and use the existing async downloader.
- The sync probe is limited to the cheap `_resolved` marker check and only updates scheduler-owned state.
- Adjusted the scheduler dedupe unit test to avoid depending on whether the local machine happens to have `react@18.2.0` in cache.

### Validation

- `cargo fmt`
- `cargo test -p utoo-pm`

### Bench plan

Trigger linux/npmjs phase bench and compare mainly:

| phase | expected signal |
| --- | --- |
| p3 cold install | should be neutral; cache misses still download through the same queue |
| p4 warm link | target phase; expect lower ctx from skipping per-package cache-hit worker tasks |

### DIFF
--- a/crates/pm/src/service/install_scheduler.rs
+++ b/crates/pm/src/service/install_scheduler.rs
@@ -10,7 +10,7 @@ use utoo_ruborist::progress::{BuildEvent, EventReceiver};
 use crate::util::cloner::{clone_count, clone_package_from_cache};
 use crate::util::downloader::{
     download_bytes, download_stats, extract_to_cache, is_registry_tarball_url,
-    registry_cache_lookup, resolve_seeded_cache_path,
+    registry_cache_lookup, registry_cache_lookup_sync, resolve_seeded_cache_path,
 };
 use crate::util::user_config::get_manifests_concurrency_limit_sync;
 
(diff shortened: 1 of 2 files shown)


elrrrrrrr:
### GHA run 1 read

Run: https://github.com/utooland/utoo/actions/runs/26015795276

| phase | utoo wall | utoo ctx | same-run utoo-next | same-run bun | read |
|---|---:|---:|---:|---:|---|
| p3_cold_install | 6.76s ±1.69 | 123.5K / 80.8K | 7.28s, 123.9K / 50.8K | 6.93s, 5.4K / 7.2K | wall ok but iCtx regresses materially |
| p4_warm_link | 2.43s ±0.15 | 50.6K / 24.0K | 2.56s, 42.1K / 19.0K | 3.51s, 266 / 25 | not positive; ctx is worse than same-run baselines and much worse than #2965 |

Conclusion: do not fold this as-is. Moving only the registry cache-hit probe into the scheduler does not reduce p4 scheduling cost; it likely just shifts the hot path while seeded-cache probe and clone worker costs remain dominant.

elrrrrrrr:

PR #2967, opened 2026-05-18
perf(pm): resolve install seeded cache synchronously

elrrrrrrr:
## Summary

- resolve ruborist-seeded install cache paths synchronously inside the install scheduler
- remove the per-package tokio task used only for cache marker probing
- keep clone/download/extract behavior unchanged for this AB

## Hypothesis

p4 warm-link still pays a per-package async cache-probe cost before falling through to the registry cache path. Moving the `_resolved` marker probe into the scheduler should reduce context switches without changing IO volume.

## Validation

- cargo fmt
- cargo test -p utoo-pm service::install_scheduler::tests
- cargo test -p utoo-pm util::downloader::tests

## Benchmark Plan

- GHA label-triggered pm-bench-phases, Linux/npmjs only
- Compare p3/p4 wall and vCtx/iCtx against #2948 baseline

### DIFF
--- a/crates/pm/src/service/install_scheduler.rs
+++ b/crates/pm/src/service/install_scheduler.rs
@@ -10,7 +10,7 @@ use utoo_ruborist::progress::{BuildEvent, EventReceiver};
 use crate::util::cloner::{clone_count, clone_package_from_cache};
 use crate::util::downloader::{
     download_bytes, download_stats, extract_to_cache, is_registry_tarball_url,
-    registry_cache_lookup, resolve_seeded_cache_path,
+    registry_cache_lookup, resolve_seeded_cache_path_sync,
 };
 use crate::util::user_config::get_manifests_concurrency_limit_sync;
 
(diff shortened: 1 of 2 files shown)


elrrrrrrr:
### GHA run 1 read

Run: https://github.com/utooland/utoo/actions/runs/26016267872

| phase | utoo wall | utoo ctx | same-run utoo-next | same-run bun | read |
|---|---:|---:|---:|---:|---|
| p3_cold_install | 10.01s ±2.35 | 115.8K / 76.6K | 10.76s, 122.7K / 56.4K | 6.84s, 7.5K / 8.7K | wall and iCtx regress; not useful |
| p4_warm_link | 2.30s ±0.12 | 47.8K / 22.6K | 2.46s, 42.3K / 18.7K | 3.38s, 377 / 29 | not positive; ctx remains high and worse than same-run utoo-next/npm |

Conclusion: do not fold this as-is. The seeded-cache marker probe alone is not the p4 lever; #2965 points to clone worker scheduling as the useful boundary. This PR also has a mac e2e failure, but the Linux perf result is already enough to reject the experiment.

elrrrrrrr:

PR #2969, opened 2026-05-18
perf(pm): use sync bin linking checks

elrrrrrrr:
## Summary

AB experiment for p3/p4 install performance.

This keeps package cloning and scheduler behavior unchanged. The only hot path moved is the bin-link rebuild path used by `utoo install --ignore-scripts`:

- replace per-bin async `try_exists().await` with `Path::try_exists()`
- add `ScriptService::ensure_executable_sync()` for bin linking
- keep existing behavior: skip missing bin targets, add shebang for text bins without one, preserve binary files, and ensure Unix executable permissions

## Hypothesis

p4 warm link has no network and uses a warm package cache, so remaining ctx should mostly come from small filesystem syscalls and per-bin async dispatch. Running these tiny bin checks synchronously should reduce scheduler/context-switch overhead without changing clone/download behavior.

## Validation

Passed:

- `cargo fmt`
- `cargo test -p utoo-pm test_ensure_executable_sync`
- `cargo test -p utoo-pm test_execute_queues_skips_missing_bin_file`

Attempted:

The full workspace clippy fails before this change is evaluated in `pack-core` because the local `next.js` symlinked checkout does not match the pack-core API usage (`CrossOrigin`, `ExternalType::Promise`, and Issue trait receiver/signature errors). PM-targeted clippy is clean.

## Benchmark Plan

Label-trigger GHA bench, compare p3/p4 against same-run `utoo-next`, `utoo-npm`, and `bun`. Keep only if p4 wall/ctx moves positively without p3 regression.

### DIFF
--- a/crates/pm/src/service/package.rs
+++ b/crates/pm/src/service/package.rs
@@ -377,7 +377,7 @@ impl PackageService {
                 }
 
                 let target_path = package.path.join(relative_path);
-                if !crate::fs::try_exists(&target_path).await? {
+                if !target_path.try_exists()? {
                     tracing::debug!(
                         "Binary file {} does not exist, skipping",
                         target_path.display()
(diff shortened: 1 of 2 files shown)


elrrrrrrr:
### GHA run 1 read

Run: https://github.com/utooland/utoo/actions/runs/26017108305

| phase | utoo wall | utoo ctx | same-run utoo-next | same-run utoo-npm | same-run bun | read |
|---|---:|---:|---:|---:|---:|---|
| p3_cold_install | 6.69s ±1.70 | 115.6K / 79.6K | 7.74s, 128.4K / 52.9K | 7.22s, 113.0K / 47.5K | 6.78s, 4.1K / 7.6K | wall is fine, but iCtx regresses materially |
| p4_warm_link | 2.32s ±0.16 | 51.1K / 25.1K | 2.42s, 38.5K / 17.2K | 2.38s, 42.2K / 18.8K | 3.34s, 405 / 25 | not acceptable; ctx is worse than same-run next/npm |

Conclusion: reject this single-factor experiment. Moving bin-link existence/chmod checks to sync does not reduce the p4 ctx problem; the small wall movement is not worth the ctx regression. Do not fold.

elrrrrrrr:

PR #2979, opened 2026-05-18
perf(pm): fast path registry cache package clones

elrrrrrrr:
## Summary

- add a registry-cache clone fast path that uses the standard `package/` wrapper directly
- keep seeded/git/file cache clones on the existing auto-discovery path
- carry cache layout through the install scheduler so workers remain primitive I/O operations

## Hypothesis

Warm registry installs currently call the generic source-discovery path before every cache clone, which costs at least one directory scan per package. Since npm registry tarballs are extracted into `<cache>/<name>/<version>/package`, the scheduler can skip that scan for registry download/cache-hit paths and reduce p4 ctx/syscall without changing non-registry cache semantics.

## Validation

- `cargo fmt`
- `CARGO_NET_OFFLINE=true cargo test -p utoo-pm util::cloner::tests::test_clone_registry_package_from_cache_sync_uses_package_dir`
- `CARGO_NET_OFFLINE=true cargo test -p utoo-pm service::install_scheduler::tests`

## Benchmark Result

GHA run: https://github.com/utooland/utoo/actions/runs/26025415891

npmjs phase bench:

| phase | wall | vCtx | iCtx | note |
| --- | ---: | ---: | ---: | --- |
| p0 | 7.56s ±1.69 | 109.8K | 58.9K | noisy |
| p1 | 2.28s ±0.01 | 17.3K | 20.1K | unchanged target |
| p3 | 4.64s ±0.36 | 70.9K | 44.5K | no clear ctx win vs #2975 |
| p4 | 1.74s ±0.08 | 14.1K | 9.2K | wall good, vCtx not better than #2975 best |

Conclusion: this does not currently prove a ctx/syscall win. #2975 already had p4 around `13.1-13.7K / 9.3-9.9K`; this branch is `14.1K / 9.2K`. Keep as rejected/hold unless repeated runs show otherwise.

### DIFF
--- a/crates/pm/src/service/install.rs
+++ b/crates/pm/src/service/install.rs
@@ -24,6 +24,8 @@ use crate::util::linker::link;
 use crate::util::logger::{
     PROGRESS_BAR, finish_progress_bar, log_progress, print_install_counts, start_progress_bar,
 };
+use crate::util::registry::{REGISTRY_NPMJS, REGISTRY_NPMMIRROR};
+use crate::util::user_config::get_registry;
 use utoo_ruborist::compat::{is_cpu_compatible, is_os_compatible};
 
 use super::binary::update_package_binary;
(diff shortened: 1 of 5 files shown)


elrrrrrrr:

PR #2981, opened 2026-05-18
perf(pm): cap install clone workers

elrrrrrrr:
## Summary

- change install clone worker limit from `available_parallelism * 2` to CPU parallelism, clamped to `2..=8`
- keep extract worker limit unchanged
- add a unit test for the clone concurrency cap

## Hypothesis

Warm-link p4 is clone-worker dominated. The current clone limit oversubscribes rayon workers (`n * 2`, up to 16), which can improve throughput on some disks but may inflate vCtx/iCtx and sys time on GHA. Capping clone workers to CPU parallelism should reduce scheduler churn. This AB checks whether ctx falls without giving back wall time.

## Validation

- `cargo fmt`
- `CARGO_NET_OFFLINE=true cargo test -p utoo-pm service::install_scheduler::tests::clone_concurrency_tracks_cpu_parallelism`

## Benchmark Result

GHA run: https://github.com/utooland/utoo/actions/runs/26026777144

npmjs phase bench:

| phase | wall | vCtx | iCtx | note |
| --- | ---: | ---: | ---: | --- |
| p0 | 7.35s ±0.07 | 82.9K | 48.1K | noisy but not target |
| p1 | failed | 12.5K | 16.5K | network EOF, ignore |
| p3 | 5.70s ±0.21 | 71.0K | 42.8K | roughly neutral vs #2975 |
| p4 | 2.16s ±0.05 | 15.8K | 12.2K | worse than #2975 |

Conclusion: rejected. Reducing clone worker concurrency increases p4 wall and context switches. The current `2x CPU` cap appears better for GHA warm-link behavior, likely because clone work is mostly blocking filesystem syscalls rather than CPU compute.

### DIFF
--- a/crates/pm/src/service/install.rs
+++ b/crates/pm/src/service/install.rs
@@ -24,6 +24,8 @@ use crate::util::linker::link;
 use crate::util::logger::{
     PROGRESS_BAR, finish_progress_bar, log_progress, print_install_counts, start_progress_bar,
 };
+use crate::util::registry::{REGISTRY_NPMJS, REGISTRY_NPMMIRROR};
+use crate::util::user_config::get_registry;
 use utoo_ruborist::compat::{is_cpu_compatible, is_os_compatible};
 
 use super::binary::update_package_binary;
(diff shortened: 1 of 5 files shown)


elrrrrrrr:

PR #2982, opened 2026-05-18
perf(pm): skip install progress off tty

elrrrrrrr:
## Summary

- add logger helpers that no-op progress length/inc updates when stderr is not a TTY
- route PM install/package progress updates through those helpers
- make `finish_progress_bar` no-op off TTY as well

## Hypothesis

`ProgressReceiver` already avoids indicatif work off TTY, but PM install still called `PROGRESS_BAR.inc()` directly for every package placement. In CI and `2>&1 | tee`, indicatif rendering is hidden, but each update still mutates progress state under its internal lock. Warm-link p4 has thousands of package placements, so this may show up as lock contention, vCtx/iCtx, and sys time.

This AB checks whether removing hidden progress-bar state mutation lowers p3/p4 ctx without changing local TTY UX.

## Validation

- `cargo fmt`

## Benchmark Result

GHA run 1: https://github.com/utooland/utoo/actions/runs/26031609818
GHA run 2: https://github.com/utooland/utoo/actions/runs/26032374627

npmjs phase bench:

| run | phase | wall | vCtx | iCtx | note |
| --- | --- | ---: | ---: | ---: | --- |
| 1 | p3 | 6.63s ±1.86 | 71.8K | 43.1K | noisy |
| 1 | p4 | 2.23s ±0.01 | 13.0K | 9.2K | best ctx tier |
| 2 | p3 | 5.72s ±0.28 | 62.8K | 42.0K | improved vs recent runs |
| 2 | p4 | 2.18s ±0.01 | 13.1K | 9.3K | reproduced ctx tier |

Compared with #2975 p4 `13.1-13.7K / 9.3-9.9K`, this branch is consistently at the low end: `13.0-13.1K / 9.2-9.3K`. Wall is neutral/slower on these globally slow runners, but the ctx/sys signal reproduces and the code fixes the existing inconsistency where resolver progress is TTY-gated while install progress still mutates hidden indicatif state.

Conclusion: keep as a small ctx/syscall cleanup candidate and consider folding into the main PR.

### DIFF
--- a/crates/pm/src/service/install.rs
+++ b/crates/pm/src/service/install.rs
@@ -24,6 +24,8 @@ use crate::util::linker::link;
 use crate::util::logger::{
     PROGRESS_BAR, finish_progress_bar, log_progress, print_install_counts, start_progress_bar,
 };
+use crate::util::registry::{REGISTRY_NPMJS, REGISTRY_NPMMIRROR};
+use crate::util::user_config::get_registry;
 use utoo_ruborist::compat::{is_cpu_compatible, is_os_compatible};
 
 use super::binary::update_package_binary;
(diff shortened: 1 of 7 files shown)


elrrrrrrr:

PR #2984, opened 2026-05-18
perf(pm): serialize scheduler extract writes

elrrrrrrr:
## Summary

- keep async standalone tarball extraction on the existing parallel file-write path
- make the install scheduler's sync extraction write files serially inside each tarball
- avoid nested rayon fan-out when package-level extract workers are already running concurrently

## Hypothesis

p3 cold install already has package-level extraction concurrency from the install scheduler. Each package extract worker currently enters `extract_tarball_sync`, which then fans out per-file writes through rayon again. That nested parallelism can oversubscribe the filesystem path: many package workers all enqueue file-write chunks at once.

Serial file writes inside one tarball should reduce scheduler pressure and sys time while preserving cross-package extraction parallelism.

## Validation

- `cargo fmt`
- `CARGO_NET_OFFLINE=true cargo test -p utoo-pm util::downloader -- --nocapture`

## GHA Benchmark

Runs:

- https://github.com/utooland/utoo/actions/runs/26042006409
- https://github.com/utooland/utoo/actions/runs/26043785334

linux/npmjs phase bench, this PR column only:

| phase | run 1 | run 2 |
| --- | ---: | ---: |
| p0 wall | 9.91s | 9.95s |
| p0 vCtx/iCtx | 90.0K / 56.8K | 90.8K / 51.1K |
| p1 wall | 2.56s | 2.82s |
| p1 vCtx/iCtx | 14.9K / 18.4K | 16.5K / 19.5K |
| p3 wall | 6.34s | 5.83s |
| p3 vCtx/iCtx | 69.1K / 46.5K | 67.6K / 42.5K |
| p4 wall | 2.14s | 2.16s |
| p4 vCtx/iCtx | 13.4K / 9.7K | 13.1K / 9.3K |

N=2 comparison points:

- p4 is neutral and stable, as expected, because warm-link does not extract tarballs.
- p3 remains healthy and beats the same-run `utoo-next` / `utoo-npm` baselines on wall and vCtx.
- The isolated contribution of serial per-tarball writes is still modest/uncertain because this PR is stacked on the install scheduler changes; compared with the previous stacked PR's historical GHA samples, p3 is in the same band rather than a clear step-change.

## Benchmark Plan

Use the normal GHA linux/npmjs phase bench. The expected signal is p3 sys/ctx improvement. p4 should be neutral because warm-link does not extract tarballs.

This PR is stacked on #2982 so it can be reviewed and benchmarked independently from the progress fast path.

### DIFF
--- a/crates/pm/src/util/extractor.rs
+++ b/crates/pm/src/util/extractor.rs
@@ -39,7 +39,11 @@ pub fn extract_and_write_sync(gzip_bytes: Bytes, dest: &Path) -> Result<()> {
         .with_context(|| format!("Failed to create destination directory: {}", dest.display()))?;
 
     let estimated_size = estimate_uncompressed_size(&gzip_bytes);
-    extract_tarball_sync(gzip_bytes, estimated_size, dest)
+    // The install scheduler already parallelizes at the package level.
+    // Keeping each tarball's file writes serial avoids nested rayon fan-out
+    // where many package extract workers all enqueue per-file write chunks at
+    // the same time.
+    extract_tarball_sync(gzip_bytes, estimated_size, dest, FileWriteMode::Serial)
 }
 
 /// Estimate uncompressed size from gzip footer (last 4 bytes store original size mod 2^32).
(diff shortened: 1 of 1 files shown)


elrrrrrrr:

PR #2986, opened 2026-05-18
perf(pm): clone from extracted file index

elrrrrrrr:
## Summary

Stacked on #2984. This tests the p3 hypothesis that cold install pays avoidable cache-walk cost immediately after extracting a tarball.

- return an optional file index from sync registry extraction
- deduplicate duplicate tar entries in the extracted file index while preserving first-seen order
- store that index in the install scheduler cache entry for freshly extracted tarballs
- materialize clones from the extracted index when available, avoiding a second `read_dir` walk of the package cache
- keep warm-cache and seeded-cache paths on the existing clone implementation (`index: None`)

## Expected Impact

`p3_cold_install` currently does download -> extract-to-cache -> clone-from-cache. For freshly extracted registry tarballs, extract already parsed every file path, but clone walks the cache again to rediscover those paths. Reusing the extracted file index should reduce p3 filesystem syscalls and context switches without changing cache layout or p4 behavior.

This is an AB experiment after #2985 showed demand download priority was not a p3 win.

## Benchmark Notes

Public GHA npmjs run (`pm-bench-phases`, linux, ant-design):

- `p3_cold_install`: `utoo 5.87s`, `69.0K / 46.1K ctx` vs `utoo-npm 7.03s`, `111.2K / 52.2K ctx` and `utoo-next 9.39s`, `135.1K / 48.7K ctx`
- `p4_warm_link`: `utoo 2.04s`, `13.4K / 9.7K ctx` vs `utoo-npm 2.27s`, `41.1K / 19.1K ctx`
- `p1_resolve`: stays positive from the scheduler stack, `utoo 2.39s`, `15.2K / 20.2K ctx` vs `utoo-npm 3.34s`, `77.7K / 94.2K ctx`

`npmmirror.com` did not emit output in this GHA run.

## Validation

- `cargo fmt`
- `CARGO_NET_OFFLINE=true cargo clippy -p utoo-pm --all-targets -- -D warnings --no-deps`
- `CARGO_NET_OFFLINE=true cargo test -p utoo-pm service::install_scheduler -- --nocapture`
- `CARGO_NET_OFFLINE=true cargo test -p utoo-pm util::extractor -- --nocapture`
- `CARGO_NET_OFFLINE=true cargo test -p utoo-pm util::downloader -- --nocapture`
- `CARGO_NET_OFFLINE=true cargo test -p utoo-pm util::cloner::hardlink_clone_tests -- --nocapture` (0 tests on mac; Linux-only indexed hardlink test is covered by GHA)

PR #2987, opened 2026-05-18
perf(pm): cap install tarball downloads

elrrrrrrr:
## Summary

Stacked on #2986. This AB experiment caps install tarball download concurrency independently from manifest concurrency.

- keep the resolver/manifest concurrency setting unchanged
- cap install tarball downloads at 128
- leave extract and clone worker limits unchanged

## Hypothesis

The install scheduler currently reuses manifest concurrency for tarball downloads. In p3, tarball downloads are followed by extract-to-cache and clone/materialize work, so excessive download fan-out can increase IO pressure and scheduler churn without improving end-to-end wall time. A lower tarball cap may keep the pipeline saturated while reducing contention.

## Validation

- `cargo fmt`
- `CARGO_NET_OFFLINE=true cargo test -p utoo-pm service::install_scheduler -- --nocapture`

### DIFF
--- a/crates/pm/src/service/install_scheduler.rs
+++ b/crates/pm/src/service/install_scheduler.rs
@@ -260,7 +260,7 @@ impl SchedulerState {
             done_tx,
             done_rx,
             shutdown: false,
-            download_limit: get_manifests_concurrency_limit_sync().max(1),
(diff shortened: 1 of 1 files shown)


elrrrrrrr:

PR #2989, opened 2026-05-19
perf(pm): batch install clone completions

elrrrrrrr:
## Summary

- Batch install clone completions so the scheduler receives one completion message for a small group of clone jobs instead of one per package.
- Keep clone materialization on rayon workers; the scheduler only dispatches work and records completion state.
- Use a bounded clone worker count derived from the existing clone concurrency limit, with a small fixed batch size to avoid long serialized hardlink runs.
- Integrate #2991: prioritize parent clone unblockers and send them as single-job batches so nested dependency placement is not delayed behind unrelated hardlink work in the same batch.

## Why

p3/p4 are dominated by materialization pressure after the earlier install scheduler changes. Batching clone completions reduces scheduler wakeups, but parent packages are on the nested placement critical path. If a parent clone completes early inside a batch but its completion is reported only after other leaf clones finish, its children cannot enter the queue. The #2991 follow-up fixes that without disabling batching for ordinary leaf clones.

## Validation

- `cargo fmt`
- `cargo test -p utoo-pm install_scheduler`
- `cargo clippy -p utoo-pm --all-targets -- -D warnings --no-deps`

Note: full workspace `cargo clippy --all-targets -- -D warnings --no-deps` is currently blocked in this worktree by pack-core/next.js submodule API mismatch, unrelated to the PM scheduler change.

## Benchmark

Latest #2989 reference before #2991 integration, GHA Linux npmjs run `26089858682`:

| phase | utoo wall | utoo vCtx/iCtx |
| --- | ---: | ---: |
| p0_full_cold | 7.48s | 66.5K / 43.2K |
| p1_resolve | 2.37s | 15.2K / 18.2K |
| p3_cold_install | 5.29s | 53.0K / 32.9K |
| p4_warm_link | 2.23s | 7.8K / 4.5K |

#2991 AB before integration, GHA Linux npmjs run `26093544293`:

| phase | utoo wall | utoo vCtx/iCtx |
| --- | ---: | ---: |
| p3_cold_install | 5.21s | 48.4K / 34.3K |
| p4_warm_link | 1.86s | 7.6K / 4.4K |

Integrated #2989 head (`fd38cfff`), GHA Linux npmjs run `26094747554`:

| phase | utoo wall | utoo vCtx/iCtx |
| --- | ---: | ---: |
| p0_full_cold | 7.28s | 67.4K / 44.2K |
| p1_resolve | 2.39s | 14.5K / 19.7K |
| p3_cold_install | 5.30s | 52.1K / 35.0K |
[...]

PR #2991, opened 2026-05-19
perf(pm): prioritize clone unblockers

elrrrrrrr:
## Summary

This is a p3/p4 scheduler experiment on top of #2989.

#2989 batches clone completions to reduce scheduler wakeups, but that can delay parent package completion inside a batch. A parent clone completion is on the materialize critical path because it releases `blocked_by_parent` children. This PR keeps ordinary leaf clones batched, while prioritizing ready parent unblockers and sending them as a single-job batch.

## Hypothesis

- p4 warm link should improve or stay neutral if nested placements enter the queue earlier.
- p3 cold install may improve when extract/download work and clone materialization overlap, because clone critical-path parents no longer wait behind unrelated hardlink work in the same batch.
- ctx may rise slightly if many parent clones become single-item batches; this is the tradeoff to validate.

## Validation

- `cargo fmt`

## Bench Result

GHA Linux npmjs run `26093544293`:

|---|---:|---:|---:|
| p3 cold install | 5.21s | 48.4K | 34.3K |
| p4 warm link | 1.86s | 7.6K | 4.4K |

Compared with #2989 latest npmjs reference (`p3 5.29s, 53.0K/32.9K`; `p4 2.23s, 7.8K/4.5K`), this is directionally positive. p4 is the strongest signal because this PR directly changes warm materialize scheduling.

Conclusion: keep as a candidate; the hypothesis that batched parent clone completions delay nested placement is supported.

### DIFF
--- a/crates/pm/src/service/install_scheduler.rs
+++ b/crates/pm/src/service/install_scheduler.rs
@@ -496,19 +496,7 @@ impl SchedulerState {
     }
 
     fn spawn_clone_batch(&mut self) -> bool {
-        let mut batch = Vec::new();
-        while batch.len() < CLONE_BATCH_LIMIT {
-            let Some(job) = self.clone_queue.pop_front() else {
-                break;
-            };
-            let target = job.spec.target.clone();
-            if self.clone_done.contains(&target) || !self.clone_active.insert(target.clone()) {
-                continue;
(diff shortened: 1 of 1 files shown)


elrrrrrrr:

PR #2993, opened 2026-05-19
perf(pm): reserve install clone capacity

elrrrrrrr:
## Summary

This is an independent p3 scheduler experiment on top of #2989.

Extract/cache population and clone/materialize both run blocking filesystem work on the global rayon pool. On a 2-core runner the current limits can queue extract workers and clone workers at the same time, so clone/materialize may still compete with cache writes even after completion batching.

This PR keeps full extract concurrency when there is no clone work, but halves the effective extract limit while clone work is queued or active.

## Hypothesis

- p3 may improve if clone/materialize is currently delayed by extract/cache-write rayon work.
- p4 should be neutral because warm link has no extract queue.
- If p3 regresses, extract/cache population is the critical path and this reservation is too conservative.

## Validation

- `cargo fmt`

## Bench Result

GHA Linux npmjs run `26093744012`:

|---|---:|---:|---:|
| p3 cold install | 4.84s | 59.0K | 39.2K |
| p4 warm link | 1.54s | 8.5K | 4.4K |

The p3 wall is faster than #2989 latest npmjs reference (`5.29s`), but ctx is worse (`53.0K/32.9K` -> `59.0K/39.2K`). Also, p4 should be a no-op control for this PR and moved a lot, which means this run had substantial GHA variance.

Conclusion: inconclusive. This needs a rerun or poollab confirmation before keeping; the wall signal is interesting, but p4 control noise prevents attributing it to the extract-slot reservation.

### DIFF
--- a/crates/pm/src/service/install_scheduler.rs
+++ b/crates/pm/src/service/install_scheduler.rs
@@ -464,8 +464,19 @@ impl SchedulerState {
         self.extract_queue.len() + self.extract_active.len()
     }
 
+    fn effective_extract_limit(&self) -> usize {
+        if self.clone_queue.is_empty() && self.clone_workers_active == 0 {
+            self.extract_limit
+        } else {
+            // Extract and clone both run blocking filesystem work on rayon.
+            // Once materialize has ready work, keep new cache writes from
+            // filling every extract slot and delaying clone progress.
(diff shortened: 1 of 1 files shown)


elrrrrrrr:

PR #2994, opened 2026-05-19
perf(pm): flush clone unblockers early

elrrrrrrr:
## Summary

This is a p3/p4 scheduler experiment on top of #2989 after #2991 integration.

#2991 proved that parent clone completions are on the nested placement critical path. It handles that by prioritizing parent unblockers and putting them in a single-job batch. This PR keeps that early parent signal but avoids dedicating a whole worker batch to it:

- parent unblockers are still selected first;
- clone workers send `CloneProgress` immediately after a parent unblocker finishes;
- `CloneProgress` completes that target and wakes nested children, but does not release the worker slot;
- remaining leaf clones continue in the same worker and are reported by the final `CloneBatch`.

## Hypothesis

- p4 may improve over #2989/#2991 if parent children can be queued early while leaf clones remain batched.
- ctx may rise slightly because parent progress adds a message, but worker utilization should be better than single-job parent batches.
- p3 should be neutral to slightly positive; cold extract/cache work is still the larger p3 cost.

## Validation

- `cargo fmt`

## Bench Result

GHA Linux npmjs run `26097117399`:

| phase | utoo wall | utoo vCtx/iCtx |
| --- | ---: | ---: |
| p0_full_cold | 7.40s | 71.2K / 46.7K |
| p1_resolve | 3.37s | 15.8K / 20.2K |
| p3_cold_install | 5.38s | 54.4K / 35.0K |
| p4_warm_link | 2.12s | 7.4K / 4.4K |

Reference #2989 integrated head (`fd38cfff`):

| run | p3 | p4 |
| --- | ---: | ---: |
| `26094747554` | 5.30s, 52.1K / 35.0K | 2.07s, 7.5K / 4.5K |
| `26095851789` | 5.66s, 54.7K / 37.6K | 2.16s, 7.8K / 4.6K |

Conclusion: inconclusive / do not integrate for now. The early-flush implementation is logically reasonable, but it does not beat #2991 clearly enough to justify the extra completion state and worker-slot semantics.

### DIFF
--- a/crates/pm/src/service/install_scheduler.rs
+++ b/crates/pm/src/service/install_scheduler.rs
@@ -106,6 +106,12 @@ struct ReadyClone {
     cache: RegistryCacheEntry,
 }
 
+struct CloneBatchJob {
+    target: PathBuf,
+    ready: ReadyClone,
+    flush_on_complete: bool,
+}
+
 struct DownloadedPackage {
     package: PackageFetch,
     bytes: Bytes,
@@ -133,6 +139,10 @@ enum OpDone {
     Extract {
         key: String,
         result: Result<RegistryCacheEntry, String>,
+    },
+    CloneProgress {
+        target: PathBuf,
+        result: Result<(), String>,
     },
     CloneBatch {
         results: Vec<(PathBuf, Result<(), String>)>,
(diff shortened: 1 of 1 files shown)


elrrrrrrr:

PR #2995, opened 2026-05-19
perf(pm): clone first waiter after extract

elrrrrrrr:
## Summary

This is a p3 scheduler experiment on top of #2989.

When a tarball download completes, the scheduler already knows which clone placements are waiting for that package. Today extract completes, reports cache readiness to the scheduler, then the scheduler queues clone/materialize work onto rayon. This PR moves the first ready clone waiter into the extract worker:

- download completion removes one primary clone waiter from `download_waiters`;
- extract worker writes the cache and immediately materializes that primary target;
- extract completion still records the cache and queues remaining waiters normally;
- if extract fails, the primary and remaining waiters receive the same error;
- if primary clone fails after cache succeeds, remaining waiters can still use the cache.

## Hypothesis

- p3 may improve by removing one scheduler/rayon handoff per downloaded package with a ready waiter.
- p4 should be neutral because warm link does not use the extract path.
- ctx may move either way: fewer clone tasks, but longer extract workers.

## Validation

- `cargo fmt`

## Bench Results

Baseline #2989 integrated head (`fd38cfff`):

- run `26094747554`: p3 `5.30s`, ctx `52.1K/35.0K`; p4 `2.07s`, ctx `7.5K/4.5K`
- full-green run `26095851789`: p3 `5.66s`, ctx `54.7K/37.6K`; p4 `2.16s`, ctx `7.8K/4.6K`

This PR #2995 (`d1acae9`, run `26100667673`):

- p0 `7.71s`, ctx `77.8K/50.2K`
- p1 `2.52s`, ctx `17.5K/22.0K`
- p3 `5.75s`, ctx `59.7K/33.9K`
- p4 `2.09s`, ctx `7.6K/4.3K`

Conclusion: do not integrate for now. p4 is neutral as expected, but p3 does not beat the #2989+#2991 baseline and vCtx increases. The extra coupling between extract and first materialize is not justified by this result.

### DIFF
--- a/crates/pm/src/service/install_scheduler.rs
+++ b/crates/pm/src/service/install_scheduler.rs
@@ -109,6 +109,12 @@ struct ReadyClone {
 struct DownloadedPackage {
     package: PackageFetch,
     bytes: Bytes,
+    primary_clone: Option<CloneSpec>,
+}
+
+struct ExtractOutcome {
+    cache: RegistryCacheEntry,
+    primary_clone: Option<(PathBuf, Result<(), String>)>,
 }
 
 type CloneResponder = oneshot::Sender<Result<(), String>>;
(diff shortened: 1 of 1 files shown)


elrrrrrrr:

PR #3002, opened 2026-05-19
perf(pm): force http1 tarball downloads

elrrrrrrr:
## Summary

- Force the PM tarball downloader reqwest client to use HTTP/1.1.
- Keep the change scoped to tarball downloads; resolver manifest HTTP clients already use the same protocol choice.

## Why

p3 cold install is now mostly cold tarball download plus cache extract/write. The resolver HTTP client already avoids HTTP/2 multiplexing because one slow response can stall unrelated manifest requests on the same connection. Tarball installs have the same shape with larger bodies, so this experiment applies the same protocol choice to the PM downloader.

## Validation

- `cargo fmt`
- `cargo test -p utoo-pm retry`

## Benchmark

Draft experiment. Add `benchmark` label and use the sticky `pm-bench-phases` result for GHA/Linux npmjs validation.

### DIFF
--- a/crates/pm/src/util/retry.rs
+++ b/crates/pm/src/util/retry.rs
@@ -27,6 +27,10 @@ impl std::error::Error for RetryableError {}
 pub fn build_dns_cached_client() -> reqwest::Client {
     client_builder()
         .connect_timeout(std::time::Duration::from_secs(5)) // TLS + TCP handshake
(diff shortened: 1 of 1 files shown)


elrrrrrrr:
Picked into #2989 as f893cbd4 after GHA/Linux npmjs validation (run 26123034198). Closing this experiment PR.

PR #3003, opened 2026-05-19
perf(pm): raise install tarball concurrency

elrrrrrrr:
## Summary

- raise the install tarball download default to the non-semver resolver default when the registry does not support semver and the user did not pass an explicit `--manifests-concurrency-limit`
- keep explicit CLI limits as the override path
- leave semver-capable registries on the existing default

## Validation

- `cargo fmt`
- `cargo test -p utoo-pm user_config`

Note: full workspace `cargo clippy --all-targets` is blocked in this worktree because `next.js` is not initialized; a temporary symlink to another worktree does not match this branch's pack-core API.

### DIFF
--- a/crates/pm/src/service/install_scheduler.rs
+++ b/crates/pm/src/service/install_scheduler.rs
@@ -14,7 +14,7 @@ use crate::util::downloader::{
     RegistryCacheEntry, download_bytes, download_stats, extract_to_cache_indexed_sync,
     is_registry_tarball_url, registry_cache_lookup_sync, resolve_seeded_cache_path,
 };
-use crate::util::user_config::get_manifests_concurrency_limit_sync;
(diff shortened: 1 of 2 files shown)


elrrrrrrr:
Closing this AB: GHA/Linux npmjs did not validate raising install tarball concurrency to 256 on top of the current HTTP/1 tarball baseline.

Compared with #2989 (`f893cbd`):
- p3 regressed: `5.34s` -> `5.59s`
- p3 ctx improved slightly: `32.1K / 24.3K` -> `27.6K / 22.9K`, but wall time is the priority here
- p0/p4 were neutral to slightly worse

Not picking this into #2989.

elrrrrrrr:
Closed: benchmark AB did not validate this change on GHA/Linux npmjs.

PR #3006, opened 2026-05-19
perf(pm): tune npmmirror install downloads

elrrrrrrr:
## Summary

Experimental follow-up to #2989. Keep npmjs install tarball concurrency at the existing default, but use a higher install download floor for registry.npmmirror.com.

Rationale: the global 128 experiment was rejected because GHA/npmjs regressed slightly, while the mirror environment benefits from the larger active download window. This branch scopes the change to the mirror registry so GHA/npmjs should stay neutral.

## Validation

- cargo fmt --check
- cargo test -p utoo-pm npmmirror_registry_uses_higher_install_download_floor


elrrrrrrr:
Picked into #2989 as `0b2deaad`. Public GHA/Linux npmjs run `26130251370` was neutral: p0 `7.56s`, p1 `2.38s`, p3 `5.25s`, p4 `2.15s` with npmjs still on the existing download window. Closing this experiment PR.

PR #3007, opened 2026-05-19
perf(pm): cap tokio runtime workers

elrrrrrrr:
## Summary

Experimental follow-up to #2989. Cap the PM tokio runtime worker count at 4 while leaving rayon extract/clone workers based on available parallelism.

## Why

poolab p3 measurements show much higher voluntary context switches than GHA. A 4-core affinity run reduced voluntary ctx substantially without changing wall time, but it also constrained rayon and converted part of the signal into involuntary ctx. This draft isolates the async runtime side only.

## Validation

- cargo fmt --check
- cargo test -p utoo-pm test_cli_debug_assert

## Expected Signal

GHA/npmjs should be neutral if the runner already exposes 4 or fewer effective CPUs. poolab should tell us whether high-core quota runtime polling is the main source of the p3 ctx amplification.

### DIFF
--- a/crates/pm/src/main.rs
+++ b/crates/pm/src/main.rs
@@ -50,6 +50,10 @@ use crate::helper::workspace::init_project_root;
 
 use crate::helper::migrate::{FromPm, migrate_from_pnpm};
 
+// Keep async network polling compact on high-core machines; CPU-heavy extract
+// and clone work still scales through rayon's pool.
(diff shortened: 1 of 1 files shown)


elrrrrrrr:
Closing this experiment early: the isolated tokio-worker cap did not show a clear local target signal, so it is not worth spending more CI cycles. The p3 ctx investigation should continue on download/extract scheduling rather than a global runtime cap.

PR #3008, opened 2026-05-20
perf(pm): queue manifest store writes

elrrrrrrr:
## Summary

- Queue `DiskManifestStore` writes through one dedicated background writer thread instead of spawning one Tokio task per manifest write.
- Keep resolver hot paths fire-and-forget while avoiding async runtime / blocking-pool contention from manifest persistence.
- Bound the opportunistic writer queue; if disk stalls and the queue fills, new cache writes are dropped instead of growing resolver memory without limit.
- Stream JSON serialization through `BufWriter` instead of materializing a full `Vec<u8>` before each disk write.
- Flush queued manifest writes when the store is dropped, with a unit test covering versions and per-version manifest writes.

## Why

The p1 investigation showed `versions.json` / version-manifest persistence is required for cache parity, but per-write Tokio tasks add scheduler noise on the resolve hot path. This PR keeps the same cache semantics and moves serialization + filesystem writes behind a single bounded queue. The cache is opportunistic, so dropping writes under extreme backpressure is preferable to unbounded memory growth.

## Validation

- `cargo fmt`
- `cargo test -p utoo-pm manifest_store`

Note: full workspace `cargo clippy --all-targets -- -D warnings --no-deps` is currently blocked in this worktree by pack-core / next.js submodule API mismatch, unrelated to this PM-only change.

### DIFF
--- a/crates/pm/src/util/json.rs
+++ b/crates/pm/src/util/json.rs
@@ -1,6 +1,9 @@
+use std::fs::File;
+use std::io::{self, BufWriter, Write};
 use std::path::Path;
 
 use anyhow::{Context, Result};
+use serde::Serialize;
 use serde::de::DeserializeOwned;
 
 /// Read and parse a JSON file into the specified type.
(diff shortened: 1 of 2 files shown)


elrrrrrrr:
Addressed Gemini review feedback in 48969db0:

- Replaced the unbounded writer channel with a bounded `sync_channel` and `try_send`; if the opportunistic cache queue is full, new writes are dropped instead of blocking resolution or growing memory without bound.
- Switched manifest serialization from `serde_json::to_vec` + `fs::write` to `serde_json::to_writer` through `BufWriter<File>`, with the missing-parent retry re-serializing after `create_dir_all`.

Validation: `cargo fmt`, `cargo test -p utoo-pm manifest_store`, `cargo clippy -p utoo-pm --all-targets -- -D warnings --no-deps`.

killagu (approved):
LGTM

PR #2937, opened 2026-05-12
perf: experiment main-loop bfs resolver

elrrrrrrr:
## Purpose

Integration / measurement branch for the new PM resolver and install scheduling model. The branch validates the architecture and benchmark result before splitting into reviewable PRs.

Core direction:

- main resolver loop owns BFS ordering, per-run caches, waiters, failures, and inflight dedupe;
- spawned tasks only perform concrete work: registry HTTP, JSON parse/extract on CPU pool, download/extract, or clone/link IO;
- speculative preload/prefetch must not block demand BFS work;
- npmjs on GHA needs more request fan-out than a single reqwest client pool provides.

### Resolver

- Native resolver path is reorganized around a main-loop BFS scheduler.
- Demand requests are prioritized over speculative prefetches.
- Prefetch concurrency is throttled while demand work is pending.
- Completed fetches are batch-drained so one wakeup can unblock multiple waiters.
- Registry fetch tasks return data back to the loop instead of mutating graph/global resolver state.

### Manifest data-plane

- Manifest HTTP can return raw bytes without parsing.
- Existing parsed APIs remain compatibility wrappers.
- Native `simd_json` parsing stays off the async runtime and uses global `rayon::spawn`.
- A dedicated `MANIFEST_PARSE_POOL` was tested and removed: it did not show consistent wall-time benefit, while it adds another CPU scheduler and weakens global rayon work stealing.
- Full-manifest parsing preserves raw bytes for later on-demand version extraction.

### Install

- Install scheduler owns download inflight/done/waiter state and clone inflight/done/parent ordering.
- Download/extract and clone tasks only perform IO.
- Pipeline workers no longer need to know OnceMap/global scheduler state.

### npmjs fan-out

- Native registry HTTP uses a 4-client round-robin reqwest pool.
- WASM keeps a single browser-backed client.
- Resolver manifest concurrency auto-raises to 256 for non-semver npmjs when the CLI did not explicitly set `--manifests-concurrency-limit`.
- Tarball download concurrency remains unchanged.

## Rayon Pool AB Conclusion

Conclusion: keep manifest parsing on global rayon. Do not keep a dedicated manifest parse pool.

GHA comparison, `ant-design`, `registry.npmjs.org`, `ubuntu-latest`, `N=2`:

| phase | #2937 dedicated pool (`7f9b3cbb`) | AB global rayon (`5f7817df`) | read |
|---|---:|---:|---|
[...]

PR #3012, opened 2026-05-20
perf(pm): add resolver manifest provider boundary

elrrrrrrr:
## Summary

- Add a `ManifestProvider` boundary for resolver drivers: jobs describe full-manifest fetch, version-manifest fetch, and full-manifest extraction.
- Implement the boundary for `UnifiedRegistry` while keeping existing `RegistryClient` resolve behavior unchanged.
- Factor version-manifest store dedupe into a shared helper and cover the new provider store-hit / extract persistence paths with unit tests.

## Split Plan Context

This is the first small PR for the #2937 / #2948 resolver architecture split. It only introduces the provider seam; follow-up PRs can move BFS ordering, per-run cache, waiters, and inflight de-duplication into the resolver main loop without making spawned jobs aware of global scheduling state.

Next reviewable slices:

1. Move deps resolver BFS scheduling onto the main loop using this provider boundary.
2. Move resolver-local inflight / waiter ownership out of `UnifiedRegistry`.
3. Apply the same scheduler shape to install once the deps path is stable.

## Expected Metrics

- Runtime performance: expected neutral. This PR introduces a boundary but keeps the existing resolver scheduling and cache behavior intact.
- p0/p1/p3/p4 wall: within normal noise; no improvement required.
- ctx/iCtx: no material increase expected. Spawn ownership and inflight location are not changed yet.
- Behavioral bytes: net/cache/node_modules/lock output should remain equivalent.
- Code metric: provider jobs become the reviewable seam for the upcoming main-loop scheduler PRs.

## Validation

- `cargo fmt`
- `cargo test -p utoo-ruborist provider_ --lib`

### DIFF
--- a/crates/ruborist/src/service/mod.rs
+++ b/crates/ruborist/src/service/mod.rs
@@ -49,6 +49,7 @@ pub(crate) mod fetch;
 mod fs;
 pub(crate) mod http;
 pub(crate) mod manifest;
+mod provider;
 mod registry;
 mod store;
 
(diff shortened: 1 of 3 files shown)


elrrrrrrr:

PR #3014, opened 2026-05-20
perf(pm): prepare resolver graph placement helpers

elrrrrrrr:
## Summary

- Extract resolved dependency graph placement into `place_resolved_dependency`.
- Extract BFS result event emission into `emit_process_result`.
- Keep the existing sequential preload + BFS behavior unchanged.
- Add a unit test covering node placement, resolved dependency edge state, and child edge expansion.

## Split Plan Context

This is the second PR in the resolver split stack and builds on #3012. It prepares the builder for the demand-BFS main loop by separating graph mutation from registry fetching. The next PR can dispatch manifest provider jobs from the BFS driver and reuse this placement path when jobs complete.

Base branch: `perf/pm-resolver-provider-boundary`

## Expected Metrics

- Runtime performance: expected neutral.
- p0/p1/p3/p4 wall: within normal noise; no improvement required for this preparatory refactor.
- ctx/iCtx: no material increase expected.
- Code metric: reduce duplicated graph placement/event logic before introducing main-loop scheduling.

## Validation

- `cargo fmt`
- `cargo check -p utoo-ruborist`
- `cargo test -p utoo-ruborist test_place_resolved_dependency_adds_node_and_child_edges --lib`
- `cargo test -p utoo-ruborist --lib`
- `cargo clippy --all-targets -- -D warnings --no-deps`

PR #3015, opened 2026-05-20
perf(pm): share resolved dependency placement

elrrrrrrr:
## Summary

- Add `process_dependency_with_resolved` so resolver drivers can place an already resolved package through the same graph mutation path.
- Add `reuse_existing_dependency` to share the edge-marking and node-type propagation path.
- Keep existing sequential BFS behavior unchanged; `process_dependency` now resolves first, then delegates the placement/reuse decision to the shared helper.
- Add a unit test covering resolved-package reuse without creating duplicate nodes.

## Split Plan Context

This PR builds on #3014. It is still a preparatory resolver split: the next PR can let a demand-BFS main loop receive completed manifest jobs and call `process_dependency_with_resolved` instead of duplicating placement logic.

Base branch: `perf/pm-resolver-main-loop`

## Expected Metrics

- Runtime performance: expected neutral.
- p0/p1/p3/p4 wall: within noise.
- ctx/iCtx: no material increase expected.
- Code metric: one graph placement/reuse path for sequential BFS and the upcoming main-loop driver.

## Validation

- `cargo fmt`
- `cargo test -p utoo-ruborist test_process_dependency_with_resolved_reuses_existing_node --lib`
- `cargo test -p utoo-ruborist --lib`

### DIFF
--- a/crates/ruborist/src/resolver/builder.rs
+++ b/crates/ruborist/src/resolver/builder.rs
@@ -363,6 +363,32 @@ fn place_resolved_dependency(
     ProcessResult::Created(new_index)
 }
 
+fn reuse_existing_dependency(
+    graph: &mut DependencyGraph,
+    parent: NodeIndex,
+    edge: &DependencyEdgeInfo,
+    existing: NodeIndex,
+) -> ProcessResult {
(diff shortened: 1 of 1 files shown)


elrrrrrrr:

PR #3016, opened 2026-05-20
perf(pm): add resolver demand main loop

elrrrrrrr:
## Summary

This PR is the first scheduler PR in the resolver stack. It replaces the separate preload + sequential BFS resolver path with a demand-driven BFS main loop:

- main loop owns BFS traversal, per-run manifest cache, waiters, and inflight de-duplication
- provider tasks only execute concrete manifest jobs (`Full`, `Version`, `ExtractVersion`)
- resolver-owned manifest cache now seeds from and exports to `ProjectCacheData`
- `service/api` no longer pre-populates/export manifests through `UnifiedRegistry` memory cache
- test mock registry implements `ManifestProvider` so resolver tests exercise the new boundary

This PR is intentionally larger than the preparatory PRs because the scheduler, waiters, and project-cache ownership have to move together to preserve cache behavior. Follow-up PRs should stay smaller: priority/preload fanout, speculative extract, install scheduler, extract/clone worker lanes.

## Expected metrics

| Phase | Expected result |
| --- | --- |
| p1 deps, npmjs | Primary target. ctx/ictx should drop materially versus preload + sequential BFS because main loop owns inflight and dispatches demand jobs directly. Target range: close to prior best `~15K-25K ctx`; wall should move toward `~1.8-2.2s` before store-write tuning. |
| p1 deps, npmmirror | Should improve or stay near prior best. Semver registries use direct version jobs with main-loop dedupe. |
| p0/p3/p4 install | Expected mostly neutral in this PR. Install scheduler, extract worker, and clone/materialize changes are intentionally left for follow-ups. |
| net/cache/node_mod bytes | Should remain behaviorally equivalent. Manifest project-cache output is still produced through `ProjectCacheData`, now from resolver-owned state. |
| risk guard | No poollab/internal data should be posted to GitHub comments; use GHA Linux/npmjs bench label first. |

## Validation

- `cargo fmt`
- `cargo check -p utoo-ruborist`
- `cargo test -p utoo-ruborist test_manifest_state --lib`
- `cargo test -p utoo-ruborist --lib` (169 passed)
- `cargo clippy --all-targets -- -D warnings --no-deps`
- `cargo check -p utoo-wasm --target wasm32-unknown-unknown`

## Stack plan

1. Provider boundary (#3012)
2. Graph placement helpers (#3014)
3. Shared resolved placement (#3015)
4. Demand BFS main loop (this PR)
5. Priority/preload fanout and speculative extract
6. Install scheduler providerization
[...]

PR #3017, opened 2026-05-20
perf(pm): prioritize resolver prefetch jobs

elrrrrrrr:
## Summary

Follow-up to the demand BFS main loop. This PR adds a two-lane manifest queue:

- demand jobs always pop before prefetch jobs
- completed version manifests enqueue transitive registry deps as low-priority prefetch
- prefetch concurrency is capped at `concurrency / 4` so speculative work does not occupy the whole resolver lane
- no speculative full-manifest extraction yet; that remains a follow-up PR

## Expected metrics

| Phase | Expected result |
| --- | --- |
| p1 deps, npmjs | Should improve wall modestly when BFS discovers transitive deps, without increasing ctx materially. Demand requests should stop being delayed behind preload-style work. |
| p1 deps, npmmirror | Expected small wall improvement or neutral. Direct semver version jobs now fan out transitive prefetches after completion. |
| ctx/ictx | Should remain near #3016 or slightly better; prefetch cap should prevent ctx from climbing back toward preload levels. |
| p0/p3/p4 install | Expected neutral; install/extract/clone scheduling is still untouched. |

## Validation

- `cargo fmt`
- `cargo test -p utoo-ruborist test_fetch_queues --lib`
- `cargo test -p utoo-ruborist test_manifest_state_dedupes --lib`
- `cargo test -p utoo-ruborist --lib` (170 passed)

## Stack plan

1. Provider boundary (#3012)
5. Priority/preload fanout (this PR)
6. Speculative extract / 304 exact-version refinement
7. Install scheduler providerization
8. Extract/clone/materialize worker lane split
9. Cleanup old scheduler/cache ownership paths

### DIFF
--- a/crates/ruborist/src/resolver/builder.rs
+++ b/crates/ruborist/src/resolver/builder.rs
@@ -17,7 +17,7 @@
 
 use futures::stream::{FuturesUnordered, StreamExt};
 use petgraph::graph::NodeIndex;
-use std::collections::{HashMap, HashSet, VecDeque};
+use std::collections::{HashMap, VecDeque};
 use std::path::PathBuf;
 use std::sync::Arc;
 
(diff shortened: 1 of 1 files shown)


elrrrrrrr:

PR #3018, opened 2026-05-20
perf(pm): speculatively extract full manifest versions

elrrrrrrr:
## Summary

Follow-up to the demand/prefetch resolver queue. This PR lets a full-manifest provider job carry the spec that triggered it, then parse the full manifest and optionally extract the matching version manifest in the same off-runtime worker.

- `ManifestJob::Full` now carries `spec: Option<String>`
- `ManifestFullData::Full` can include `speculative: Option<(spec, CoreVersionManifest)>`
- full-manifest parse uses `parse_full_manifest_with_core_off_runtime` so the current version can be extracted without scheduling a second `ExtractVersion` job
- main loop stores the speculative manifest under both the original spec and exact version keys, then schedules transitive prefetches from it

## Expected metrics

| Phase | Expected result |
| --- | --- |
| p1 deps, npmjs | Small but measurable wall/ctx improvement on cold full-manifest paths by avoiding one extra extract job for the edge that triggered the full fetch. |
| p1 deps, npmmirror | Mostly neutral; semver registries already use version jobs directly. |
| ctx/ictx | Should stay flat or decrease slightly. The work is moved into an existing parse worker, not a new spawned job. |
| p0/p3/p4 install | Expected neutral; install/extract/clone scheduling remains unchanged. |

## Validation

- `cargo fmt`
- `cargo test -p utoo-ruborist test_full_fetch_result_stores_speculative_manifest --lib`
- `cargo test -p utoo-ruborist --lib` (171 passed)
- `./e2e/utoo-pm.sh` (local full e2e passed)

## Stack plan

1. Provider boundary (#3012)
5. Priority/preload fanout (#3017)
6. Speculative full-manifest extract (this PR)
7. Install scheduler providerization
8. Extract/clone/materialize worker lane split
9. Cleanup old scheduler/cache ownership paths

### DIFF
--- a/crates/ruborist/src/resolver/builder.rs
+++ b/crates/ruborist/src/resolver/builder.rs
@@ -460,7 +460,7 @@ enum FetchKey {
 impl ManifestJob {
     fn key(&self) -> FetchKey {
         match self {
-            Self::Full { name } => FetchKey::Full(name.clone()),
+            Self::Full { name, .. } => FetchKey::Full(name.clone()),
             Self::Version { name, spec, .. } | Self::ExtractVersion { name, spec, .. } => {
                 FetchKey::Version(name.clone(), spec.clone())
             }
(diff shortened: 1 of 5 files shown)


elrrrrrrr:

PR #3019, opened 2026-05-20
perf(pm): remove obsolete resolver preload

elrrrrrrr:
## Summary

- Remove the old `resolver::preload` module that is no longer called after the demand-BFS main loop moved manifest scheduling into `builder.rs`.
- Drop the public module export so new resolver work cannot accidentally route around the main-loop scheduler.
- Keep progress event variants intact for compatibility; this PR only removes the dead preload implementation and tests.

## Expected Metrics

| Phase | Expected result |
| --- | --- |
| p1 deps | Runtime neutral. The module was not on the active resolver path after #3016. |
| p0/p3/p4 install | Runtime neutral. Install scheduling is untouched. |
| ctx/iCtx | No material change expected. This is cleanup only. |
| code size | Removes 346 lines of obsolete preload implementation/tests, reducing the old scheduler surface before install-side PRs. |

## Validation

- `cargo fmt`
- `cargo test -p utoo-ruborist --lib` (167 passed)

## Stack plan

1. Provider boundary (#3012)
5. Priority/preload fanout (#3017)
6. Speculative full-manifest extract (#3018)
7. Remove obsolete preload module (this PR)
8. Install scheduler providerization
9. Extract/clone/materialize worker lane split
10. Cleanup old scheduler/cache ownership paths


elrrrrrrr:

PR #3020, opened 2026-05-20
perf(pm): stream install clone completions

elrrrrrrr:
## Summary
- drain per-depth install clone tasks with `FuturesUnordered` instead of awaiting join handles in spawn order
- keep clone execution on the existing tokio tasks; this PR only changes completion polling order

## Expected Metrics
| Phase | Expected impact | Rationale |
| --- | --- | --- |
| p3 cold install | small wall-time improvement or neutral | later completed clone tasks no longer wait behind an earlier slow join handle before the depth can finish |
| p4 warm/materialize | small wall-time improvement or neutral | same completion-order drain helps deep node_modules layers with uneven package clone times |
| ctx / iCtx | neutral | this PR does not reduce task count yet; it only removes spawn-order head-of-line waiting |
| p0 / p1 resolve | neutral | resolver flow and manifest scheduling are untouched |

## Validation
- `cargo fmt`
- `cargo check -p utoo-pm`
- `cargo test -p utoo-pm service::install::tests`

## Split Plan Position
This is the first install-side micro split after the resolver stack. Follow-up PRs should move install inflight ownership into the scheduler, then split download/extract/clone lanes separately.

### DIFF
--- a/crates/pm/src/service/install.rs
+++ b/crates/pm/src/service/install.rs
@@ -1,6 +1,7 @@
 use crate::util::cli_enum::ScriptPolicy;
 use anyhow::Context;
 use anyhow::Result;
+use futures::stream::{FuturesUnordered, StreamExt};
 use std::collections::{HashMap, HashSet};
 use std::path::Path;
 use std::time::Instant;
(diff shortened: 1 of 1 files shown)


elrrrrrrr:

PR #3021, opened 2026-05-20
perf(pm): split install cache primitives

elrrrrrrr:
## Summary
- add `resolve_seeded_cache_path` so install scheduling can distinguish seeded git/file/http cache hits from registry download fallthrough
- add `download_and_extract_to_cache` as a downloader primitive without global single-flight ownership
- add `clone_package_from_cache` as a cloner primitive for already-resolved cache paths
- keep existing `download_to_cache` and `clone_package_once` APIs as the current global OnceMap wrappers

## Expected Metrics
| Phase | Expected impact | Rationale |
| --- | --- | --- |
| p0 / p1 resolve | neutral | resolver code is untouched |
| p3 cold install | neutral in this PR | existing OnceMap/semaphore wrappers still own scheduling; this PR only creates reusable atomic operations |
| p4 warm/materialize | neutral | clone behavior and cache validation remain the same |
| ctx / iCtx | neutral | task topology is unchanged until the follow-up scheduler PR consumes these primitives |

## Validation
- `cargo fmt`
- `cargo check -p utoo-pm`
- `cargo test -p utoo-pm util::`

## Split Plan Position
This is the utility-boundary PR before moving install inflight ownership into the scheduler. It keeps the current global wrappers in place so the behavior change can be reviewed separately from scheduler integration.

### DIFF
--- a/crates/pm/src/util/cloner.rs
+++ b/crates/pm/src/util/cloner.rs
@@ -54,6 +54,25 @@ pub async fn wait_clone_if_pending(target_path: &str) {
         .await;
 }
 
+/// Clone a package from an already-resolved cache path without touching the
+/// global clone/download single-flight maps.
+pub async fn clone_package_from_cache(
+    name: &str,
+    version: &str,
+    tarball_url: &str,
+    cache_path: &Path,
+    target_path: &Path,
+) -> Result<()> {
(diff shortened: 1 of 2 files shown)


elrrrrrrr:

PR #3023, opened 2026-05-20
perf(pm): route pipeline clones through scheduler

elrrrrrrr:
## Summary
- Route pipeline download and clone events through the install scheduler instead of spawning per-event download/clone tasks.
- Share one scheduler between pipeline prefetch and install materialization so install waiters reuse scheduler in-flight state.
- Move parent clone ordering into the scheduler and remove the obsolete util-level wait_clone_if_pending helper.

## Expected Metrics
- p3 stale-lock/pipeline install: lower ctx/iCtx by replacing per-event spawned download/clone tasks and util OnceMap waits with scheduler dispatch.
- p4 warm/fresh-lock install: expected neutral vs #3022; same scheduler path remains active.
- p0/p1 resolver: expected neutral except less pipeline-side task pressure while resolve emits placement events.

## Observed GHA Bench
- `bench-phases-linux` passed on run 26182235197.
- npmjs p1: `utoo` 2.86s, vCtx/iCtx 31.0K/36.6K; `utoo-next` 2.86s, 48.8K/71.9K.
- npmjs p3: `utoo` 6.54s, vCtx/iCtx 104.5K/54.7K; `utoo-next` 7.52s, 130.6K/49.8K.
- npmjs p4: `utoo` 2.09s, vCtx/iCtx 44.4K/22.3K; `utoo-next` 2.25s, 41.3K/19.4K.
- npmmirror: no output captured by this GHA run.

## Split Plan Position
- Builds on #3022 scheduler core.
- Next PR: remove or narrow remaining install util OnceMap fallback surface once CI/bench confirms scheduler coverage.

## Validation
- cargo fmt
- cargo check -p utoo-pm
- cargo test -p utoo-pm service::install_scheduler
- cargo test -p utoo-pm service::pipeline
- cargo test -p utoo-pm service::install::tests
- cargo clippy --all-targets -- -D warnings --no-deps
- GHA: clippy / format / builds / utoo-pm CI / linux+mac e2e / bench-phases-linux passed; Windows e2e still pending at last check

PR #3024, opened 2026-05-20
perf(pm): remove clone oncemap fallback

elrrrrrrr:
## Summary
- Make install materialization always use the install scheduler.
- Remove the obsolete clone_package_once / CLONE_CACHE OnceMap fallback from cloner.
- Keep downloader cache and git resolver caches unchanged; those still serve non-install helper paths.

## Expected Metrics
- p3/p4 install materialization: lower lock/context overhead from removing the remaining clone OnceMap fallback surface.
- p0/p1 resolver: expected neutral.
- Overall: should be mostly code simplification; performance should not regress relative to #3023.

## Split Plan Position
- Builds on #3023 after pipeline events route through scheduler.
- This is the cleanup PR that removes the old clone single-flight path once scheduler coverage is complete.

## Validation
- cargo fmt
- cargo check -p utoo-pm
- cargo test -p utoo-pm service::install_scheduler
- cargo test -p utoo-pm service::pipeline
- cargo test -p utoo-pm service::install::tests
- cargo test -p utoo-pm util::cloner

### DIFF
--- a/crates/pm/src/service/install.rs
+++ b/crates/pm/src/service/install.rs
@@ -68,10 +68,8 @@ pub async fn install_packages(
     groups: &HashMap<usize, Vec<(String, Package)>>,
     cwd: &Path,
     omit: &HashSet<OmitType>,
-    clone_scheduler: Option<&InstallScheduler>,
+    clone_scheduler: &InstallScheduler,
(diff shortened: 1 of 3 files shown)


elrrrrrrr:

PR #3025, opened 2026-05-20
perf(pm): optimize version manifest hot path

elrrrrrrr:
## Summary

Completes the missing p1 hot-path pieces from the original resolver performance branch:

- read exact-version manifest responses directly into a mutable `Vec<u8>` buffer
- parse version manifests from that buffer with `simd_json` off the async runtime
- remove the provider-level `DashSet` write-dedupe set so cache writes stay in the host `ManifestStore` writer queue instead of adding resolver hot-path shared state

## Expected performance signal

This is intended to recover the original main-PR p1 ctx shape. #2948 reported p1 around `1.79s` with `15.7K/19.6K` v/iCtx; #3024 was still around `2.72s` with `27.9K/36.6K`, so this PR should primarily move p1 ctx down.

## Validation

- `cargo fmt`
- `cargo test -p utoo-ruborist service::manifest`
- `cargo test -p utoo-ruborist service::registry`

### DIFF
--- a/crates/ruborist/src/service/manifest.rs
+++ b/crates/ruborist/src/service/manifest.rs
@@ -15,11 +15,13 @@ use super::http::get_client;
 use crate::model::manifest::{CoreVersionManifest, FullManifest};
 use crate::resolver::version::resolve_target_version;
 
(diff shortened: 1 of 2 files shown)


elrrrrrrr:

PR #3026, opened 2026-05-20
perf(pm): split install download and extract lanes

elrrrrrrr:
## Summary

Completes the missing p3 install scheduling pieces from the original performance branch:

- split install tarball download and extract/write into separate scheduler lanes
- keep download slots focused on cache lookup + network bytes
- bound extract work independently and queue clones only after cache materialization finishes
- tighten install clone lane fanout to reduce clone/materialize contention

## Expected performance signal

This should recover the p3 shape from the original scheduler experiments by preventing download tasks from holding their slot through extraction and file writes. It is stacked after the version-manifest hot-path PR so the final stack can be compared directly with #2948.

## Validation

- `cargo fmt`
- `cargo check -p utoo-pm`
- `cargo test -p utoo-pm service::install_scheduler`
- `cargo test -p utoo-pm util::downloader`

### DIFF
--- a/crates/pm/src/service/install_scheduler.rs
+++ b/crates/pm/src/service/install_scheduler.rs
@@ -2,11 +2,14 @@ use std::collections::{HashMap, HashSet, VecDeque};
 use std::path::{Path, PathBuf};
 
 use anyhow::{Context, Result, anyhow};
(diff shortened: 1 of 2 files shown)


elrrrrrrr:

PR #3028, opened 2026-05-21
perf(pm): source resolver mainloop architecture

elrrrrrrr:
## Summary

Draft source PR for the resolver half of #2948. This is not the final review unit; it is the source branch used to verify the full resolver stack in one place.

This source branch now matches the current resolver split-stack top (`perf/pm-split-resolver-doc-cleanup`), so the split PRs should compose back to this diff exactly.

## Covers From #2948

- `ManifestProvider` / `ManifestJob` boundary.
- Resolver-owned BFS/cache/inflight/waiter state.
- Provider jobs only perform concrete manifest work.
- Demand-first prefetch scheduling.
- Obsolete preload removal.
- Speculative full-manifest version extraction.
- Version manifest Vec hot path.
- Ruborist global `MemoryCache` / `PackageCache` / registry OnceMap cleanup.
- PM resolver wiring needed to preserve p1 behavior.

## Actual Split Stack

- #3030 `perf(pm): add resolver manifest provider boundary` (+84)
- #3031 `perf(pm): parse version manifests from vec buffers` (+75 / -11)
- #3034 `perf(pm): extract requested core from full manifest parse` (+65 / -15)
- #3035 `perf(pm): add mock manifest provider` (+101)
- #3036 `perf(pm): fan out registry http clients` (+38 / -4)
- #3038 `perf(pm): add resolver pm wiring` (+115 / -113)
- #3039 `perf(pm): execute resolver manifest provider jobs` (+269)
- #3040 `perf(pm): share resolver placement helpers` (+101 / -49)
- #3041 `perf(pm): require manifest provider for builder` (+34 / -10)
- #3042 `perf(pm): thread warm project cache into resolver config` (+12 / -2)
- #3043 `perf(pm): add resolver demand mainloop` (+981 / -313)
- #3049 `test(pm): cover resolver demand mainloop` (+469 / -1)
- #3044 `perf(pm): remove obsolete resolver preload` (-366)
- #3045 `perf(pm): remove registry cache hook` (-20)
- #3046 `perf(pm): make unified registry stateless` (+151 / -493)
- #3047 `perf(pm): remove resolver memory cache layer` (+13 / -223)
- #3048 `perf(pm): update resolver cache ownership docs` (+5 / -5)

## Notes

[...]

PR #3034, opened 2026-05-21
perf(pm): extract requested core from full manifest parse

elrrrrrrr:
## Summary

Review-sized split from #3028 / #2948, stacked after #3031.

Adds the full-manifest parse helper used by the upcoming provider implementation:

- keeps `parse_full_manifest_off_runtime` as the compatibility wrapper;
- adds `parse_full_manifest_with_core_off_runtime` to parse the full manifest and optionally extract the current edge's core version in the same worker;
- preserves `FullManifest::raw` for later on-demand extraction;
- keeps native parsing on rayon and wasm parsing inline;
- adds a regression test for raw byte preservation and speculative `latest` extraction.

This PR does not change resolver scheduling yet; it only exposes the CPU helper consumed by the provider/job PR.

## Size

- Diff against #3031: `1 file changed, 65 insertions(+), 15 deletions(-)`

## Validation

- `cargo fmt`
- `cargo test -p utoo-ruborist service::manifest::tests::parse_full_manifest_with_core_extracts_requested_spec`

`pack-napi` still warns locally because `next.js` is a symlink in this worktree; clippy exits successfully.

### DIFF
--- a/crates/ruborist/src/service/manifest.rs
+++ b/crates/ruborist/src/service/manifest.rs
@@ -13,6 +13,7 @@ use super::fetch::{
 };
 use super::http::get_client;
 use crate::model::manifest::{CoreVersionManifest, FullManifest};
+use crate::resolver::version::resolve_target_version;
 
 /// Parse a JSON buffer on rayon's CPU thread pool (native) or inline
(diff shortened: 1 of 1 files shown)


elrrrrrrr:

PR #3037, opened 2026-05-21
perf(pm): remove obsolete install pipeline

elrrrrrrr:
## Summary
- remove the old install pipeline module after scheduler wiring takes over download/clone orchestration
- keep this PR deletion-only so the scheduler behavior is reviewed in the previous PR

## Verification
- cargo fmt
- cargo check -p utoo-pm


elrrrrrrr:

PR #3050, opened 2026-05-21
perf(pm): route stale installs through scheduler

elrrrrrrr:
## Summary
- route stale-lock dependency resolution events into the install scheduler
- remove the old compiled downloader/cloner OnceMap fallback now that scheduler owns inflight dedupe
- update HTTP tarball install docs to reference scheduler-owned cache resolution

## Verification
- cargo fmt
- cargo check -p utoo-pm
- cargo test -p utoo-pm service::install_scheduler -- --nocapture


elrrrrrrr:

PR #3051, opened 2026-05-21
perf(pm): schedule fresh lock installs

elrrrrrrr:
## Summary
- add the install scheduler core for fresh-lock clone materialization
- route fresh-lock installs through scheduler-owned download/extract/clone dedupe
- keep stale-lock resolver pipeline unchanged in this step so the split stays reviewable

## Verification
- cargo fmt
- cargo check -p utoo-pm


elrrrrrrr:

PR #3029, opened 2026-05-21
perf(pm): source install scheduler architecture

elrrrrrrr:
## Summary

Draft source PR for the installer half of #2948. This is not the final review unit; it is the source branch used to verify the full installer stack in one place.

This source branch now matches the current installer split-stack top (`perf/pm-split-installer-scheduler-core`), so the split PRs should compose back to this diff exactly.

## Covers From #2948

- Primitive downloader cache operations.
- Sync cloner primitive for rayon workers.
- InstallScheduler-owned download/extract/clone inflight and waiter state.
- Fresh-lock install scheduling.
- Stale-lock resolver event scheduling.
- Pipeline worker/receiver cleanup.
- Clone/download OnceMap fallback removal from the install hot path.
- HTTP tarball install docs updated for scheduler-owned cache resolution.

## Actual Split Stack

- #3032 `perf(pm): split downloader cache primitives` (+91 / -38)
- #3033 `perf(pm): add sync cloner primitive` (+225 / -79)
- #3051 `perf(pm): schedule fresh lock installs` (+490 / -5)
- #3052 `test(pm): cover install scheduler core` (+94)
- #3050 `perf(pm): route stale installs through scheduler` (+220 / -270)
- #3037 `perf(pm): remove obsolete install pipeline` (-258)

## Notes

The scheduler core is split so each PR has a clear review focus: primitives, fresh-lock scheduling, tests, stale-lock event wiring, and final deletion of the old pipeline module. The production PRs avoid dead-code staging and do not use clippy `allow` escapes.

## Validation

Validated at installer split-stack top:

- `cargo fmt`
- `cargo check -p utoo-pm`
- `cargo test -p utoo-pm service::install_scheduler -- --nocapture` (`4 passed`)
- `cargo clippy --all-targets -- -D warnings --no-deps`

`pack-napi` warns locally because `next.js` is a symlink in this worktree; clippy exits successfully.

PR #3070, opened 2026-05-25
perf(pm): split install cache primitives

elrrrrrrr:
## Summary
- Split downloader into cache lookup, byte download, and extract primitives.
- Add sync cloner primitives for scheduler workers.

## Review Focus
- Primitive APIs are stateless and do not own global single-flight state.
- **No retry inside the sync clone primitives.** `clone_dir_native_sync` (both platforms) is a single attempt that returns the error as-is; whether/how to retry is the caller's decision, not baked into the util layer. The legacy async `clone` keeps its existing `Retry::spawn` (out of scope here); only the new sync path is affected.

## Follow-up (deferred to a standalone cleanup PR)
Conclusion from review, recorded here so the cleanup can continue later:

- **Install hot path is sync-only.** The `InstallScheduler` owns dedup/inflight/waiter and calls the sync primitives on its worker pool; `clone_package_once` + the clone/download `OnceMap` single-flight are removed downstream (see #3029).
- **The async `clone_package` / `clone` stay only for the single-package path in `helper/lock.rs`**, which already runs in an async context (resolve → download → clone, serially) and does not need the scheduler.
- **Keep two entry points, but not two implementations.** The async `clone` still duplicates the platform clonefile/hardlink logic and still wraps it in `Retry::spawn`. The async entry should later collapse into a thin `spawn_blocking` wrapper over the stateless sync core — removing the duplicate, pushing the retry decision out to the caller, and fixing `clonefile` running as a blocking syscall on the async runtime (macOS).
- **This convergence is intentionally NOT done in this PR**, so the split stack composes back to #3029 exactly. It belongs in a cleanup PR after the scheduler stack lands and the hot path is stable.

### DIFF
--- a/crates/pm/src/util/cloner.rs
+++ b/crates/pm/src/util/cloner.rs
@@ -4,6 +4,7 @@ use std::sync::atomic::{AtomicUsize, Ordering};
 use anyhow::{Context, Result};
 use once_cell::sync::Lazy;
 use serde::de::DeserializeOwned;
+use tokio::task::spawn_blocking;
 use tokio_retry::Retry;
 use utoo_ruborist::manifest::IdentityView;
 use utoo_ruborist::util::OnceMap;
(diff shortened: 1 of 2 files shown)


elrrrrrrr:
Superseded: installer work is consolidated into #3029 (perf/pm-source-installer) and will be re-split from there. The review fixes from this PR (harden transient I/O, stateless sync primitives, top-use imports) are folded into #3029's cleanup commits.

PR #3071, opened 2026-05-25
perf(pm): add install scheduler state

elrrrrrrr:
## Summary
- Add InstallScheduler state machine for download/extract/clone de-duplication.
- Add coverage for fetch dedupe, extract handoff, clone dedupe, and parent clone barriers.

## Review Focus
- State machine invariants, parent-child clone ordering, and worker error propagation.


elrrrrrrr:
Superseded: the installer scheduler architecture + cleanup now lives directly in #3029 (perf/pm-source-installer) and will be re-split from there. Closing this draft to avoid a stale parallel split stack.

PR #3078, opened 2026-05-25
perf(pm): wire resolver demand driver + retire preload

elrrrrrrr:
Top of the resolver demand-loop rework. The **driver** that coordinates the store (#3079) and the scheduler (#3080), plus the registry/builder rework.

- `demand/driver.rs` — `run_main_loop_bfs` + the fetch orchestration (`schedule` / `apply` / transitive `prefetch`), operating on a separate `ManifestState` (store) and `FetchQueues` (scheduler) that stay unaware of each other; `ResolutionMode` lives here. Single-threaded, kept saturated to `concurrency` via `pump_fetches` (always N in flight, never more).
- `registry.rs` — `UnifiedRegistry` made stateless (drops in-mem cache + `OnceMap`) + impls `ManifestProvider`.
- builder wired to `demand::run_main_loop_bfs`; delete `preload.rs` + `BuildEvent::Preload*`; cache slimmed; `ProjectCacheData` gains `resolved_manifests` / `from_resolved` (persistence ↔ neutral-tuple adapter); http client-pool / concurrency / persistence tweaks.

Stacks on #3080 → #3079 → next. #3028 kept as the compose-back reference.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3079, opened 2026-05-26
perf(pm): add resolver demand manifest store

elrrrrrrr:
Part of the resolver demand-loop rework (split from the old combined #3079). The per-run manifest **store**, an independent module.

`crates/ruborist/src/resolver/demand/state.rs`:
- `ManifestSlot<K,V>` — `{ cache, waiters, failures }` per manifest kind (full keyed by name, version by `(name, spec)`), with `is_settled` / `wake`.
- `ManifestState` — the slots + `versions_cache`, with get/set accessors (`version_manifest` / `version_failure` / `is_version_settled` / `cache_version` / `fail_version` / `wake_version`), `seeded(...)`, and neutral `ResolverManifestCache` output.

Pure storage: **no fetch scheduling** (that is #3080, `queue`) and **no orchestration** (that is the driver, #3078). **Format-agnostic** — speaks neutral `(name, spec, manifest)` tuples; the `ProjectCacheData` ↔ tuples adapter lives on `ProjectCacheData` itself (in #3078), so the store never knows the on-disk format.

Bottom of the stack: #3080 (scheduler) and #3078 (driver) build on this.

### DIFF
--- a/crates/ruborist/src/resolver/demand/mod.rs
+++ b/crates/ruborist/src/resolver/demand/mod.rs
@@ -1,0 +1,13 @@
+//! Demand-driven resolver, split by concern:
+//!
+//! - [`state`] — per-run manifest store (cache, waiters, failures).
+//!
+//! The fetch scheduler ([`queue`]) and the driver that coordinates them land in
+//! the follow-up PRs; until then the store is staged here and exercised by its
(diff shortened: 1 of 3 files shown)


killagu (approved):
LGTM

PR #3083, opened 2026-05-27
perf(pm): demand resolver edge-selection module (select)

elrrrrrrr:
## What

Third leaf module of the demand resolver, after `state` (#3079) and `queue` (#3080). `select` is the **pure per-edge resolution decision**:

```rust
fn select_edge(state, edge, name, spec, mode) -> EdgeStep
// EdgeStep = Resolve | Skip | Fail | Park { wait, fetch }
```

Given the current `ManifestState` and one dependency edge, it decides the outcome: resolve from cache now, skip (optional with no match), fail (recorded fetch failure), or park the edge and enqueue a fetch.

## Why it's a clean standalone module

It only **reads** `ManifestState` and returns a decision value — no `&mut`, no async, no I/O, no graph mutation. So it's **unit-testable in isolation**, which the integration-only tests never were. 7 tests cover: semver/full-manifest cache hits, recorded failures → `Fail`, cache miss → `Park`, client-side version resolution from a cached full manifest, and optional-dep skip.

## Staging

Dead-code-staged (`#![allow(dead_code)]`), exactly like `state`/`queue` were — the driver that consumes `select_edge` lands in the follow-up PR (the demand mainloop).

### DIFF
--- a/crates/ruborist/src/resolver/demand/mod.rs
+++ b/crates/ruborist/src/resolver/demand/mod.rs
@@ -2,8 +2,9 @@
 //!
 //! - [`state`] — per-run manifest store (cache, waiters, failures).
 //! - [`queue`] — fetch scheduling, single-flight de-duplication, priority.
+//! - [`select`] — per-edge resolution decision (pure; reads the store).
 //!
-//! The two are independent; the driver that coordinates them (and exposes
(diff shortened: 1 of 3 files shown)


killagu (approved):
LGTM

fireairforce (approved):

Issue #3096, opened 2026-06-02
ut pm-pack: add `--pack-destination <dir>` to control output location


killagu:
## Summary

`ut pm-pack` always writes the tarball to the package's own directory. There is no way to redirect the output to a shared collection folder.

## Reproduction

```bash
ut pm-pack --help
# Options:
#       --dry-run        Perform a dry run without creating a tarball
#       --verbose
#       --registry <REGISTRY>
#       --cache-dir <CACHE_DIR>
#       --legacy-peer-deps
#       --manifests-concurrency-limit <…>
#       -h, --help
# (no --pack-destination)
```

To collect tarballs into one folder today you have to `cd` into each package, run `ut pm-pack`, then `mv` the result — scripted manually.

## Expected behavior

Match `npm pack --pack-destination <dir>` (also supported by pnpm):

```bash
ut pm-pack --pack-destination ./.tarballs    # writes <name>-<version>.tgz under ./.tarballs
```

## Why this matters

Downstream tooling (e2e harnesses, publish scripts, CI artifact uploads) typically expects every tgz at one known path. Without `--pack-destination` every consumer has to know where utoo placed each file and stitch a glob, which is brittle in workspace setups.

See [eggjs/egg#5959](https://github.com/eggjs/egg/pull/5959) for a real consumer that worked around this.

## Related

- workspace/catalog resolution (filed separately)
- recursive pack `-r` / `--filter` (filed separately)

Issue #3095, opened 2026-06-02
ut pm-pack: support recursive / workspace-wide pack (`-r`, `--filter`)


killagu:
## Summary

`ut pm-pack` only operates on a single package directory. There is no equivalent of `pnpm -r pack` to pack every publishable workspace package in one command.

## Reproduction

```bash
ut pm-pack -r
# error: unexpected argument '-r' found

ut -r pm-pack
# error: unexpected argument '-r' found
```

The only workaround today is a shell loop that `cd`s into each package and calls `ut pm-pack`, which has to re-implement utoo's own publishable-package discovery.

## Expected behavior

Match pnpm's pack ergonomics:

- `ut pm-pack -r` / `ut pm-pack --recursive` — pack every workspace package that isn't `private: true`
- `ut pm-pack --filter <pattern>` — same filter language as `ut install --filter`, so you can pack a subset

## Why this matters

Monorepos that produce many tarballs (publish flows, ecosystem-ci, local link testing) currently can't replace `pnpm -r pack` with a single utoo command. We had to keep a pnpm install in CI for this step alone — see [eggjs/egg#5959](https://github.com/eggjs/egg/pull/5959).

## Related

- workspace/catalog resolution: filed separately
- `--pack-destination`: filed separately

PR #3102, opened 2026-06-03
feat(pm): resolve workspace:/catalog: protocols in packed manifests

### LINKED ISSUE #3094: ut pm-pack: resolve `workspace:` and `catalog:` protocols in emitted manifests
## Summary

`ut pm-pack` emits tarballs whose `package.json` retains the literal `workspace:*` and `catalog:` strings. Anything that later runs `npm install` against the tarball (publish-to-internal-registry, e2e harnesses, `file:` overrides) fails immediately with:

```
npm error code EUNSUPPORTEDPROTOCOL
npm error Unsupported URL Type "catalog:": catalog:
```

## Reproduction

In an utoo workspace using `catalog:` (and `workspace:*` for internal deps):

```bash
cd packages/foo            # foo has `bar: workspace:*` and `baz: catalog:`
ut pm-pack
tar -xzf foo-*.tgz -C /tmp/x
cat /tmp/x/package/package.json   # still contains "bar": "workspace:*", "baz": "catalog:"
```

## Expected behavior

When packing, rewrite manifest fields the same way pnpm does:

- `workspace:*` / `workspace:^` / `workspace:~` → the workspace package's current semver (with caret/tilde semantics)
- `catalog:` / `catalog:<name>` → the version pinned in the catalog section of `pnpm-workspace.yaml`

This is the behavior of `pnpm pack` and `bun pack`; without it, monorepos that use the catalog feature cannot publish or run downstream e2e via utoo.

## Impact

I'm migrating the [eggjs/egg](https://github.com/eggjs/egg) monorepo from pnpm to utoo. The pack step was the only thing that forced us to keep `pnpm/action-setup` around in CI ([PR](https://github.com/eggjs/egg/pull/5959)) — without this fix every consumer of our prebuilt tgz files breaks.

## Related

- `ut pm-pack -r` (recursive pack) is filed separately
- `ut pm-pack --pack-destination` is filed separately

elrrrrrrr:
## Summary

Closes #3094.

`ut pm-pack` / `utoo publish` emitted tarballs whose `package.json` kept the literal `workspace:*` and `catalog:` specifiers. Any downstream `npm install` of such a tarball fails immediately with `EUNSUPPORTEDPROTOCOL`, which forced monorepos migrating to utoo (e.g. [eggjs/egg#5959](https://github.com/eggjs/egg/pull/5959)) to keep `pnpm` around **just** for the pack step.

This rewrites both protocols at pack/publish time the way pnpm/bun do:

| spec | result (member version `1.2.3`) |
|---|---|
| `workspace:*` / `workspace:` | `1.2.3` |
| `workspace:~` | `~1.2.3` |
| `workspace:^` | `^1.2.3` |
| `workspace:^1.2.0` | `^1.2.0` (prefix stripped) |
| `workspace:./path` | `1.2.3` |
| `catalog:` / `catalog:<name>` | version pinned in the catalog |

## Design

- **New** `resolve_workspace_spec` in `crates/ruborist/src/spec/mod.rs`, symmetric to the existing `resolve_catalog_spec`.
- **New** `crates/pm/src/service/publish_manifest.rs` orchestrates the rewrite:
  - No-op (returns `None`) when the manifest has neither protocol, so standalone packages are packed byte-for-byte.
  - Resolves against the **discovered workspace root** — `pm-pack` runs inside a member dir, so `workspace:` versions come from sibling members and `catalog:` from the root `.utoo.toml` (migrated from `pnpm-workspace.yaml`), not the cwd.
  - Unresolvable specifiers `bail` with an actionable message rather than shipping a broken tarball.
- `pm_pack::pack` substitutes the rewritten manifest into the tarball entry (`append_override`); the **on-disk `package.json` is never mutated** (pnpm parity).
- `publish` now builds the payload from `pack_result.manifest`, so registry metadata and tarball contents can't drift.

## Test plan

- Unit: `resolve_workspace_spec` (all forms), `rewrite_dep_specs` across all four dependency kinds, unresolvable workspace/catalog errors, protocol detection. — pm `publish_manifest` (5) + ruborist `spec` pass.
- Existing `pm_pack` (11) and full `utoo-pm` (248) suites green; `cargo fmt --check` + `cargo clippy --all-targets -D warnings` clean.
- Manual E2E: a 2-member workspace with `workspace:^`/`workspace:*` + `catalog:` → tarball `package.json` rewritten to concrete ranges, unknown fields (`repository`/`keywords`) preserved, source manifest unchanged.

## Follow-ups (out of scope)

- `workspace:<name>@<range>` aliasing (`npm:` rewrite) — rare; not handled yet.
- `ut pm-pack -r` / `--pack-destination` are tracked separately (#3095 / #3096).

### DIFF
--- a/crates/pm/src/service/mod.rs
+++ b/crates/pm/src/service/mod.rs
@@ -11,6 +11,7 @@ pub mod package;
 pub mod package_management;
 pub mod pm_pack;
 pub mod publish;
+pub mod publish_manifest;
 pub mod rebuild;
 pub mod script;
 pub mod update;
--- a/crates/pm/src/service/pm_pack.rs
+++ b/crates/pm/src/service/pm_pack.rs
@@ -12,7 +12,8 @@ use crate::model::package::PackageInfo;
 use crate::service::publish_manifest::normalize_publish_manifest;
 use crate::service::script::{ScriptOutput, ScriptService};
 use crate::util::integrity::compute_integrity;
-use crate::util::user_config::get_or_load_package_json;
+use crate::util::json::load_package_json;
+use crate::util::user_config::{get_or_load_package_json, set_package_json};
 
 #[derive(Default)]
 pub struct PackResult {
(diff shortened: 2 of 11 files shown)


killagu (approved):
LGTM

PR #2758, opened 2026-04-03
feat(pm): add private npm registry auth

elrrrrrrr:
## Summary
- Support Bearer token authentication for private npm registries
- Token resolved from `NPM_TOKEN` / `NODE_AUTH_TOKEN` env vars or `~/.utoo/config.toml` (`utoo login`)
- ruborist HTTP client made injectable (`set_http_client`) — stays auth-unaware
- Tarball downloads use hostname guard to prevent token leakage to third-party CDNs

## Design
| Layer | Responsibility |
|-------|----------------|
| **ruborist** | Zero auth awareness. HTTP client settable from outside via `set_http_client`, otherwise auto-built |
| **pm/user_config** | Resolves token, builds client with `Authorization: Bearer` default header, injects into ruborist |
| **pm/downloader** | Holds token locally, attaches Bearer only when tarball URL host matches registry host |

## Test plan
- [ ] `NPM_TOKEN=<token> ut install` with a private scoped package
- [ ] `utoo login` + `ut install` reads token from `~/.utoo/config.toml`
- [ ] Public registry install still works without token
- [ ] Tarball from a CDN URL does not receive the auth header

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3103, opened 2026-06-04
feat(pm): private npm registry auth for install/view

elrrrrrrr:
## Summary

`ut install` / `ut view` / `utoo add` against a **private registry** returned 401 — the install path (manifest fetch in ruborist, tarball download in pm) carried no credentials. This resolves a Bearer token once and attaches it, scoped to the registry host.

Supersedes #2758 (that branch predates the demand-resolver rework and the `service::auth` infrastructure; rebuilt fresh on current `next`).

## Design

**Credentials live on the registry, resolution lives on the host.**

- **ruborist** — `UnifiedRegistry` gains an `auth_token`. The provider attaches it to manifest requests (always the registry host by construction). ruborist never reads env/config — the token is *injected*.
- **`BuildDepsOptions` consolidation** — previously carried `registry_url` + `manifest_store` + `supports_semver`, and `build_deps` rebuilt the registry internally, duplicating what `Context::registry()` already assembled. Replaced those three fields with a pre-built `registry: UnifiedRegistry` the host supplies. One place builds the registry (URL + store + semver + auth); install and `resolve_package` share it.
- **pm** —
  - `auth::cached_token()` resolves the token **once** at startup (env `NPM_TOKEN`/`NODE_AUTH_TOKEN` → `~/.utoo/config.toml`), then in-memory. Matches how npm/bun load credentials once and do per-request in-memory lookups — no per-fetch disk reads.
  - `auth::token_for_url()` is the **leak guard**: a token is attached to a tarball download only when the URL host matches the registry host (npm nerf-dart semantics). bun omits this check and can leak the token to a third-party CDN that a manifest's `dist.tarball` points at — we don't.
- **wasm** — browser builds have no token source, so the registry is unauthenticated (native-only token resolution).

## Scope

Single-registry model (utoo's current `get_registry()`). Out of scope: per-scope registries (`@scope:registry=...`), npmrc `:_authToken` parsing, and basic-auth (`_auth` / user:pass) — follow-ups.

## Test plan

- Unit: `same_host` leak guard — registry tarball ✓, scheme-agnostic ✓, case-normalized ✓, third-party CDN ✗, sibling subdomain ✗, unparsable URL ✗ (fail closed).
- Existing ruborist (181) + pm (249) suites green; `cargo fmt --check`, `cargo clippy --all-targets -D warnings`, `typos` clean; wasm target compiles.
- Manual (private registry): `NPM_TOKEN=… ut install` a private scoped pkg; `utoo login` + install reads config token; public registry install still works tokenless; a CDN tarball receives **no** `Authorization` header.

### DIFF
--- a/crates/pm/src/cmd/view.rs
+++ b/crates/pm/src/cmd/view.rs
@@ -23,8 +23,10 @@ async fn view_with_registry(package_spec: &str, registry_url: &str) -> Result<()
 
     tracing::debug!("Resolved package: {name} (spec: {version_spec})");
 
-    // Fetch full manifest directly from registry (Complete format for display, no ETag)
-    let token = auth::cached_token().await;
+    // Fetch full manifest directly from registry (Complete format for display, no ETag).
+    // token_for_url applies the leak guard: a token only when registry_url is
+    // the configured registry host.
+    let token = auth::token_for_url(registry_url).await;
     let (full_manifest, _etag) = fetch_full_manifest_fresh(
         registry_url,
         name,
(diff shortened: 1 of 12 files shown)


killagu (approved):
LGTM

PR #3106, opened 2026-06-04
fix(pm): resolve global prefix correctly for npm-style installs

elrrrrrrr:
## Problem

`ut i -g <pkg>` (and `ut link`) inferred the global **prefix** from `current_exe()` using one rule for every layout: the bin dir = the executable's own directory (`exe.parent()`). That rule is correct when the binary sits directly in an on-PATH dir, but **wrong when utoo itself was installed as an npm global package**: the real binary then lives at `<prefix>/lib/node_modules/utoo/bin/ut` and the on-PATH `<prefix>/bin/ut` is just a symlink. On Linux `current_exe()` resolves the symlink to the deep path, so `exe.parent()` is utoo's **own package bin** (`.../utoo/bin`), not the real `<prefix>/bin`:

```
$ ut i -g bun        # bun installed, but...
$ which bun          # not found!
# bun landed at  <prefix>/lib/node_modules/utoo/lib/node_modules/bun
# bin symlinked  <prefix>/lib/node_modules/utoo/bin/bun   (not on PATH)
```

## Two layouts, detected by the path

The fix distinguishes two install layouts and only changes the broken one.

| Layout | Example exe | Global bin dir | Global package dir |
|--------|-------------|----------------|--------------------|
| **flat** (binary dropped directly in an on-PATH dir) | `/usr/local/bin/utoo`, `<workspace>/dist/utoo`, `target/release/utoo` | the binary's **own dir** (`exe.parent()`) — unchanged from before | `<parent-of-bin>/<GLOBAL_NODE_MODULES>` |
| **npm-style** (utoo installed as an npm package) | `<prefix>/lib/node_modules/utoo/bin/ut` | `<prefix>/bin` (Unix) / `<prefix>` (Windows) | `<prefix>/<GLOBAL_NODE_MODULES>` |

**Discriminator** — does the exe path contain a `node_modules` segment?

```rust
fn npm_style_prefix(exe: &Path) -> Option<PathBuf> {
    let components = exe.components().collect();
    // last `node_modules` segment; absent -> None == flat
    let idx = components.iter().rposition(|c| c.as_os_str() == "node_modules")?;
    // Unix global dir is `lib/node_modules`, so drop the preceding `lib` too
    let cut = if !cfg!(windows) && idx > 0 && components[idx-1] == "lib" { idx-1 } else { idx };
    Some(components[..cut].iter().map(|c| c.as_os_str()).collect()) // everything before == <prefix>
}
```

`Some(prefix)` → npm-style (climb out to the real `<prefix>`); `None` → flat (keep `exe.parent()`, which was always correct there). Works whether or not `current_exe()` resolves the bin symlink. npm always installs into a `node_modules` dir, so this precisely separates "utoo was installed as an npm package" from "the binary was dropped in a bin dir".

## Fix

- **Layout-aware inference** (above): npm-style climbs out of `lib/node_modules/<pkg>` (and any scope dir) to the real `<prefix>`; flat keeps the original `exe.parent()` behavior.
- **Windows bin dir** = the prefix **root** (npm places global `.cmd`/`.ps1` shims directly in the prefix on Windows), vs `<prefix>/bin` on Unix.
- **Prefix normalized to absolute** so a relative `UTOO_PREFIX`/config value doesn't produce cwd-relative install dirs.
- **Persistent prefix config**: `resolve_global_prefix()` with priority `--prefix` > `UTOO_PREFIX` env > `prefix` config key. Set with:
  ```
  ut config set prefix <path> --global
  ```
- **`ut link`**: link the package's bins into `<prefix>/bin` (on PATH) instead of the linked package's own `bin/` dir, matching `install -g`.

## Tests

- **Unit** (`global_bin`): flat (incl. the `dist/utoo` CI layout), npm-style, scoped, Windows-prefix-root, relative→absolute; `resolve_global_prefix` precedence.
- **e2e** — covered on **Linux, macOS (arm64 + x64), and Windows** (`utoo-pm.sh` + `utoo-pm.ps1`):
  - after `npm install -g utoo --prefix=<tmp>`, run `install -g cowsay` **through the npm-style symlink** and assert the bin lands in the prefix bin dir (root on Windows) and the package under the global node_modules — and that nothing leaks into utoo's own package dir;
  - a `UTOO_PREFIX` override case;
  - a `link-with-bin` fixture + `utoo link --prefix <tmp>` asserting the bin lands in the prefix bin dir.

All CI green across all platforms (24/24). `cargo fmt` / `cargo clippy --all-targets -- -D warnings` clean.


killagu (approved):
LGTM

PR #3109, opened 2026-06-04
fix(pm): keep cache on same device as node_modules to avoid copy fallback

elrrrrrrr:
## What

When the cache dir and the project's `node_modules` live on **different filesystems**, `hardlink`/`clonefile` can't cross the device boundary, so the cloner copies every file — no disk savings, just slower (the `WARN cross-device hardlink ... Invalid cross-device link (os error 18); falling back to copy` case).

## Changes

- **`set_cache_dir` (startup, common case):** when the **default** cache (`~/.cache/nm`) lands on a different device than the project, relocate it to a project-local cache (`node_modules/.cache/nm`) so links keep working (logged at **INFO**). An **explicitly** configured cross-device `cache-dir`/`UTOO_CACHE_DIR` is **respected** but **WARN**s that installs will copy. Generalizes the existing Windows-only drive-boundary relocation to Unix (`st_dev`) and unifies both platforms.
- **`cloner` (macOS, correctness):** `clonefile` `EXDEV`/`ENOTSUP` now falls back to hardlink/copy instead of erroring, so multi-volume layouts still install. Per-package cross-device detection stays — **no global latch** (a latch would wrongly force copy on same-device targets in a multi-volume layout).
- **e2e:** Case 8.7 forces a real cross-device split via `/dev/shm` (tmpfs, Linux only) and asserts both the **copy fallback** (explicit cache) and the **relocation** (default cache).

## Why not auto-pick a same-device cache globally

Matches bun/pnpm/npm: a single global cache maximizes cross-project reuse; fragmenting it per mount point breaks that and can't cover monorepo members spread across volumes. So: relocate only the *default* when cross-device, respect explicit config (warn), and keep the cloner's per-package copy fallback as the multi-volume safety net.

### DIFF
--- a/crates/pm/src/util/cloner.rs
+++ b/crates/pm/src/util/cloner.rs
@@ -49,7 +49,9 @@ use std::os::unix::ffi::OsStrExt;
 #[cfg(target_os = "macos")]
 use libc::clonefile;
 
-#[cfg(not(target_os = "macos"))]
+/// Hardlink-first directory clone with a copy fallback. Used directly on
+/// Linux/Windows, and on macOS as the fallback when `clonefile` can't run
+/// (non-APFS volume or cross-device).
 mod hardlink_clone {
     use std::collections::HashSet;
     use std::path::{Path, PathBuf};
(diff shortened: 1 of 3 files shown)


killagu (approved):
LGTM

PR #3123, opened 2026-06-05
fix(pack): fix wasm build and add e2e for @utoo/web

xusd320:
## Summary

- Add a Playwright e2e test for `utooweb-demo` that installs dependencies, builds the demo project, opens `dist/index.html`, and asserts the built preview renders.
- Mirror browser console output into the Playwright process with `[browser:<type>]` prefixes, and fail the test on known fatal browser messages such as `Worker: Dependency ...`, build failures, install failures, and panics.
- Add stable `data-testid` hooks to `utooweb-demo` controls, panels, file-tree nodes, and the preview iframe so the browser test can operate on stable elements.
- Update `utooweb-ci.yml` to build `@utoo/web`, install Chromium, start `utooweb-demo`, run the e2e test, and upload Playwright/demo logs on failure.
- Add the root `test:e2e:utooweb` script, Playwright dev dependency, and ignored Playwright output directories.
- Align CI/release Rust toolchains back to `nightly-2026-04-02`, including the root `rust-toolchain.toml`.
- Adjust `utooweb-release.yml` to derive the package version from the published release tag, install npm 11.6.2 for trusted publishing, and publish without mutating git tags/workspace versions.
- Update the `next.js` submodule pointer and the generated WASM binding declaration order from the current web build.

## Why

This PR now focuses on adding end-to-end browser coverage for the `@utoo/web` demo path. The test exercises the install/build/preview workflow that depends on WASM, OPFS, service workers, and loader workers, and it keeps browser console output visible in local and CI logs so failures are easier to diagnose.

The loader-worker source changes are no longer part of the current PR diff, so this description intentionally does not describe that fix.

## Validation

- `npx playwright test -c e2e/utoo-web/playwright.config.ts --list`
- `npm run build:local --workspace @utoo/web`
- `npm run start --workspace utooweb-demo -- --port 18082` plus `UTOOWEB_DEMO_URL=http://127.0.0.1:18082 npm run test:e2e:utooweb` (`1 passed`)
- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/utooweb-ci.yml"); puts "ok"'`
- `git diff --check`

Note: `npx biome check e2e/utoo-web/utooweb-demo.spec.ts` reports that `e2e/` is ignored by the current Biome configuration, so it does not process the spec file.


.github/workflows/cargo-publish.yml
```diff
@@ -46,7 +46,7 @@ jobs:
       - name: Install Rust
         uses: dtolnay/rust-toolchain@stable
         with:
-          toolchain: nightly-2026-05-15
```
yuzheng14, line 49: 为啥改回 `04-02` 了？next.js 已经更新到 `05-15` 了

xusd320: 用 `05-15` ，wasm 上构建时， `tokio` 不会调度 `future` ， 导致构建一直卡住， 还不知道啥原因

fireairforce (approved):

PR #3130, opened 2026-06-08
perf(pack): optimize transform setup

fireairforce:
## Summary

This PR trims a couple of fixed setup costs in Utoopack's module option setup:

- Skip constructing PostCSS package mappings, import maps, and transform options when there is no inline `styles.postcss` config.
- Add a cheap pre-check before running the `__webpack_public_path__` mutating AST transform, so ordinary modules without that global return early.

## Impact

Projects that do not configure PostCSS avoid paying the PostCSS transform setup cost during client/server context construction. Projects that do use inline `styles.postcss` continue to take the existing path.

The webpack public path transform still runs for modules that reference unresolved `__webpack_public_path__`; modules that do not reference it avoid the mutating visitor.

## Validation

- `cargo fmt --check`
- `cargo check -p pack-core`
- `cargo test -p pack-tests --test snapshot test_tests__snapshot__public_path__runtime -- --exact`
- `cargo test -p pack-tests --test snapshot test_tests__snapshot__style__inline_css_postcss -- --exact`

### DIFF
--- a/crates/pack-core/src/client/context.rs
+++ b/crates/pack-core/src/client/context.rs
@@ -245,13 +245,18 @@ pub async fn get_client_module_options_context(
         SourceMapsType::None
     };
     let postcss_config_content = (*config.postcss_config_content().await?).clone();
-    let postcss_package_mapping = get_postcss_package_mapping().to_resolved().await?;
-    let postcss_transform_options = Some(PostCssTransformOptions {
-        postcss_package: Some(postcss_package_mapping),
-        config_location: PostCssConfigLocation::ProjectPathOrLocalPath,
-        config_content: postcss_config_content,
-        ..Default::default()
-    });
+    let postcss_package_mapping = if postcss_config_content.is_some() {
+        Some(get_postcss_package_mapping().to_resolved().await?)
+    } else {
+        None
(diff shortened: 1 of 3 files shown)


fireairforce:
Added commit `04935276` for utoopack CLI defaults.

Summary:

- Enable persistent caching by default for `utoo pack build` and `utoo pack dev`, while preserving explicit `persistentCaching: false`.
- Apply production defaults for `concatenateModules`, `removeUnusedExports`, and `removeUnusedImports` in utoopack CLI config resolution.
- Align webpack compatibility mode with the same production optimization defaults without overriding explicit `false` values.

Validation:

- `npx tsc -p packages/pack-shared/tsconfig.json --noEmit`
- `npx tsc -p packages/pack-cli/tsconfig.json --noEmit`
- `npx tsc -p packages/pack/tsconfig.json --noEmit`
- `npx biome check packages/pack-shared/src/webpackCompat.ts packages/pack-cli/src/utils/common.ts packages/pack/src/commands/build.ts packages/pack/src/core/hmr.ts`
- pre-push hook Biome check

elrrrrrrr (approved):

PR #3151, opened 2026-06-11
perf(pm): drop redundant lockfile and package.json parsing on install

elrrrrrrr:
Split out of #3146 — single orthogonal change so the benchmark comment shows its isolated effect.

- **package-lock.json parsed once per install**: the freshness check now returns the parsed lock (`load_package_lock_if_fresh`), killing the second multi-MB serde parse in both `install()` and `ensure_package_lock()`.
- **Rebuild collect trusts the lock's `hasInstallScript` marker**: the execution queues only consume preinstall/install/postinstall — exactly what the marker covers, and the same signal the early filter already trusts. Bin-only packages and `--ignore-scripts` runs (what the bench measures) skip the per-package package.json read entirely; remaining reads run `buffer_unordered(32)` instead of one awaited syscall+parse at a time.

Expected bench signal: **p3/p4 improve modestly** (startup + post-link tail), p1 unchanged.

### DIFF
--- a/crates/pm/src/helper/lock.rs
+++ b/crates/pm/src/helper/lock.rs
@@ -16,7 +16,7 @@ use super::workspace::find_workspace_path;
 use crate::fs;
 use crate::util::cli_enum::{PackageAction, SaveType};
 use crate::util::git_resolver::{resolve_git_spec, resolve_github_spec};
-use crate::util::json::{load_package_lock_json_from_path, read_json_file};
+use crate::util::json::read_json_file;
 use crate::util::user_config::{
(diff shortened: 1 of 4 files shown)


elrrrrrrr:
关闭：核心改动是 load_package_lock_if_fresh 把 freshness 检查返回已解析的 lock，但为此引入了额外的 hash key 缓存路径，复杂度与收益不成正比（得不偿失）。p3/p4 的启动/收尾小幅收益不足以支撑这层额外状态。其余 perf 拆分（#3149 mimalloc / #3150 clone / #3152 binary-mirror）继续在 #3157 重构后的 next 上重新落地。

PR #3152, opened 2026-06-11
perf(pm): cache binary-mirror-config on disk with a 6h TTL

elrrrrrrr:
Split out of #3146 — single orthogonal change so the benchmark comment shows its isolated effect.

With any non-npmjs registry (npmmirror etc.) every install fetched `binary-mirror-config@latest` over the network — on a fully-warm install that TLS round trip was the only network left, and it blocked the first mirror-matched package's reify.

- serve from `<cache_dir>/_binary-mirror-config.json` while fresher than **6h** (worst-case staleness of a new mirror entry stays within a workday; the fetch is already best-effort)
- warm the `OnceCell` in the background at install start so a cold fetch overlaps the pipeline

Expected bench signal: visible only on the **npmmirror leg** (p4/p3); the npmjs label leg should read ✅ ~0% everywhere (config skipped on npmjs). Local A/B against npmmirror showed warm-link variance collapse (σ ±368ms → ±11ms).

### DIFF
--- a/crates/pm/src/service/binary.rs
+++ b/crates/pm/src/service/binary.rs
@@ -94,9 +94,79 @@ static CONFIG: OnceCell<BinaryMirrorConfig> = OnceCell::const_new();
 /// Cached result of whether we should skip binary mirror envs
 static SKIP_BINARY_MIRROR: OnceLock<bool> = OnceLock::new();
 
+/// On-disk cache TTL for `binary-mirror-config`. The config changes rarely;
+/// without this cache every warm install on a non-npmjs registry pays a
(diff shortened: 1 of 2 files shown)


elrrrrrrr:
Superseded by #3181, which bundles `binary-mirror-config` at build time instead of caching a runtime fetch. Since it's a package we maintain (public-registry-only, rarely changing), bundling removes the fetch entirely — no disk cache / TTL / ETag needed, and a schema drift becomes a CI failure on our side rather than a silent runtime degradation. Measured that the `/pkg/latest` endpoint has no ETag (npmmirror has no validator), so conditional-GET freshness wasn't available anyway. Closing in favor of the static bundle.

PR #3153, opened 2026-06-11
perf(pm): cut resolver CPU on the driver hot path

elrrrrrrr:
The p1 physics: metadata transfer is ~206MB on ant-design/npmjs (~1.4s at runner line rate); everything between that floor and utoo's ~2.2s is client CPU, most of it serialized on the single-threaded driver. Three structural costs removed:

1. **Speculative version extract no longer re-parses the packument.** The oneshot path cloned the multi-MB body and built a full `simd_json` borrowed tree to read one version object. Now: zero-copy shallow `RawValue` pass that brace-matches past every other version — still builds/retains no per-version index.
2. **O(1) physical-child lookup.** `find_in_parent_chain` scanned every root child (string compare each) per ancestor level, for every edge, twice (pre-fetch reuse probe + post-fetch placement — both still run; the graph changes between them). A per-parent `name → child` map maintained in `add_physical_edge` (graph is append-only) makes each probe a hash hit.
3. **Per-package memoized version selection.** When the `latest` dist-tag doesn't satisfy a spec (old-major pins like `^17.0.0`), `max_satisfying` parsed the entire version list per distinct spec — react-scale packages: ~2k parses each, on the driver. Version lists now parse + sort descending once per package (`OnceLock`); each spec walks top-down and early-exits on the first (= max) match. The latest-first shortcut is untouched.

Also per review: `workspace:` specs in a project that declares no workspaces now warn instead of vanishing into a debug-level skip; overrides already short-circuit on an empty name set.

Expected bench signal: **p1 (and p0's resolve half) improve**; p3/p4 unchanged.

### DIFF
--- a/crates/pm/src/helper/tree_builder.rs
+++ b/crates/pm/src/helper/tree_builder.rs
@@ -7,7 +7,9 @@ use std::path::PathBuf;
 
 use anyhow::Result;
 use serde_json::{Value, json};
-use utoo_ruborist::builder::{DevDeps, EdgeContext, add_edges_from, add_workspace_member};
+use utoo_ruborist::builder::{
+    DevDeps, EdgeContext, add_edges_from, add_workspace_member, resolve_workspace_member_edges,
+};
 use utoo_ruborist::graph::{DependencyGraph, EdgeType};
 use utoo_ruborist::resolver::runtime::install_runtime;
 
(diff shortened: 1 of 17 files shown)


killagu (approved):
LGTM

PR #3154, opened 2026-06-11
feat(web): inject env.read_custom_section host import for link-section

lukesandberg:
## What

The [`link-section`](https://docs.rs/link-section) crate emits a wasm host import on `wasm32-unknown-unknown`:

```
(import "env" "read_custom_section" (func (param i32 i32 i32 i32) (result i32)))
```

`wasm-bindgen` passes this import through into the module but never supplies it, so once `link-section` enters the dependency graph the `@utoo/web` wasm module fails to instantiate with a `LinkError`. This PR supplies the missing host function.

## Why now

`link-section` is pulled in transitively by the upcoming Turbopack change vercel/next.js#94503 ("migrate turbo-tasks to scattered collect"), which uses out-of-band wasm custom sections to collect registrations scattered across crates. That change hasn't landed in our `next.js` submodule yet, so this is staged ahead of it — see **Status** below.

## How

`packages/utoo-web/cli/patch-wasm-imports.js` is a post-`wasm-bindgen` codemod that injects the `env.read_custom_section` host function into the generated glue's `__wbg_get_imports()`. That's the single chokepoint both the main-thread `__wbg_init` path and the worker `initSync` path funnel through, so one patch covers every instantiation site (including the shared-module / shared-memory worker re-init).

The host function reads out-of-band custom sections — which the wasm runtime itself cannot see — via `WebAssembly.Module.customSections`, and copies the bytes into linear memory following link-section's contract (return `0` if absent; return the required length if the buffer is too small so the caller retries; otherwise copy and return the byte count). It closes lazily over the glue's `wasm.memory` and `__wbg_init.__wbindgen_wasm_module`, both populated before the wasm start function — and thus link-section's ctors — run.

It's wired into the `bindgen-dev` / `bindgen-build` / `bindgen-build:local` scripts, since `src/utoo/index.js` is git-ignored and regenerated on every build. The codemod is idempotent and throws loudly if wasm-bindgen's output shape changes, so it can't silently produce a broken build.

## Tests

`packages/utoo-web/cli/patch-wasm-imports.test.js` (`node --test`):
- source transform — injection, idempotency, loud failure on missing anchors, valid-JS output;
- the injected reader run against a **real hand-assembled wasm module carrying a custom section**, covering the size-probe / too-small / exact-copy / missing-section branches.

## Verification & status

> **Draft for handoff.** Built but not exercised end-to-end, because the motivating dependency (vercel/next.js#94503) isn't in the submodule yet and the local wasm toolchain isn't fully provisioned (no `wasm32-unknown-unknown` target / `rust-src`; pinned `wasm-bindgen` 0.2.108 vs 0.2.106 locally).

- ✅ **`wasm-opt -Oz` preserves the custom sections.** Verified empirically with binaryen v125 (the pinned version) and the build's exact flags — arbitrary named custom sections round-trip intact through the production optimize step. (The common "wasm-opt strips custom sections" warning refers to debug / `name` sections gated behind `-g`, not user sections.)
- ⬜ wasm-bindgen leaves the `env.read_custom_section` import untouched (high confidence — it passes through non-`wbg` imports).
- ⬜ the exact `link-section` version Turbopack pulls still imports from module `env` (0.18.2 does; its CHANGELOG notes this was a fix).

Both open items can only be confirmed once #94503 lands and `link-section` is actually in the graph.


fireairforce:
i cherry-pick this change into: https://github.com/utooland/utoo/pull/3191 and it now has been merged, thanks~ @lukesandberg

PR #3160, opened 2026-06-15
perf(pm): reuse package-lock.json as resolved-tree baseline, retiring the project cache

elrrrrrrr:
## Problem

`ut install <pkg>` (and any install where the manifest changed) re-resolves the **entire** dependency tree from scratch via the demand resolver. The only thing reused today is the `.utoo-manifest.json` manifest cache — the resolved *tree* is recomputed. Two consequences:

1. **Redundant full computation** — every node is re-resolved even though the existing `package-lock.json` already pins it.
2. **Tree reshuffles** — the demand resolver places nodes concurrently (128-way), and `find_compatible_node` hoisting is order-sensitive: whichever sibling resolves first gets hoisted. So the placement differs run-to-run, and adding a single leaf can redraw the whole tree — a large, noisy lock diff and a full re-clone.

`utoo install` (no args) already short-circuits to install-from-lock when nothing changed; this PR fixes the **changed-manifest** path.

## Approach — seed the resolver from the lockfile, resolve only the delta

When a usable `package-lock.json` exists, seed the resolver graph with the prior resolved tree (the inverse of `serialize_to_packages`), then let the BFS resolve only what changed:

- **Seed** (`resolver::seeding::seed_graph_from_lock`): every regular (non-root, non-link) lock entry becomes a pinned node at its locked physical path, with a synthetic `CoreVersionManifest` built from the entry; its recorded deps are seeded **already-resolved** against the slot the lock hoisted them to.
- **Resolve the delta**: importer edges (root + workspaces) stay *live/unresolved* from the current manifests. `seed_level` only enqueues unresolved edges, and `process_pending_edge` tries `try_reuse_dependency` (no network) before fetching — so unchanged deps reuse seeded nodes for free, and only added/changed deps do real work.
- **Prune** (reachability-filtered serialize): a re-resolved bump shadows the stale seeded node in the last-wins `child_index`; `graph.reachable_nodes()` (the same importer→valid-edge traversal `compute_node_types` uses) drops anything the importers no longer reach — removed deps, shadowed versions, orphaned transitives. No graph mutation, so the append-only invariant holds.

Because the lockfile pins every resolved manifest itself, it **supersedes the `.utoo-manifest.json` project cache** on this path (the user's suggestion): `project_cache` is dropped when a baseline is present. Global installs (no lock) keep using it.

Unparseable or absent locks (e.g. old `lockfileVersion: 1` with no `packages` map) fall back to a full cold resolve — the feature only ever costs a warning.

## Validation

Local, against `registry.npmjs.org` (release-local binary):

| action | result |
|---|---|
| `install debug@4.3.4` then `install is-number@7.0.0` | `debug`/`ms` **byte-identical** (version + resolved); only `node_modules/is-number` added; `0 downloaded` |
| `uninstall is-number` | prunes just `is-number` |
| `uninstall debug` | prunes `debug` **and** its transitive `ms` |

- **Unit** (`resolver::seeding`): identity fixpoint, removed-subtree prune, shared-transitive retention, shadowed-bump prune, hoist-chain resolution.
- **e2e** (`e2e/utoo-pm.sh`): new case asserting an add leaves every prior lock entry unchanged and introduces only the added subtree.
- `cargo test -p utoo-ruborist` (193) + `-p utoo-pm` (268) green; `cargo fmt` + `clippy` clean; rust-code-guard pass.

## Scope / follow-ups
- **Draft** pending review of the seeding/pruning edge cases on large real monorepos.
- Workspaces are handled by the same mechanism (importers = root + members); broad workspace soak-testing still wanted.
- Transitive `file:`-link deps are not seeded (rare); they re-resolve from the live edge if still declared.
- `@utoo/web` wasm still cold-resolves (`baseline: None`); wiring the OPFS lockfile is a follow-up. Local wasm build unverified (clang wasm sysroot unavailable) — relying on CI.


elrrrrrrr:
@codex review

fireairforce (approved):

PR #3161, opened 2026-06-15
fix(pm): route tarball cache by true source, not resolved URL

elrrrrrrr:
## Problem

`is_registry_tarball_url(url)` (in `util/downloader.rs`) classified a tarball by parsing its **resolved download URL** as a `Protocol`:

```rust
matches!(url.parse::<Protocol>(), Ok(Protocol::Http))
```

But `Protocol::Http` only checks the `https://` prefix — and **both** a registry package and a direct http(s) tarball dependency resolve to an `https://….tgz` URL. So the check cannot tell them apart; it reports every https tarball as "registry". (`Protocol` is meant for *specs* — a registry dep's spec is a semver, which parses to `Err` — but this passed it the resolved URL, the wrong layer.)

### Impact (correctness + perf)

A direct http tarball dep is normally served from its own `_http_<sha256(url)>` cache slot (seeded during resolution, URL-hashed precisely so it can't collide with a registry package). When that slot is **missing** — e.g. a **lockfile install with no prior resolve** — `resolve_seeded_cache_path` returned `Ok(None)` and the download pipeline fell through to `extract_to_cache`, which **unconditionally** writes the registry slot `<cache>/<name>/<version>/`. The http tarball's bytes then land in — and, via the shared global `~/.cache/nm`, get served for — a **registry package of the same `name@version`**. Silent cross-cache pollution.

Triggers only when an http(s) tarball dependency is present and its `_http_` slot is cold; pure-registry projects are unaffected. Not introduced here — it predates the scheduler; the misclassification was just never surfaced.

## Fix — carry the true source end-to-end

Stop re-deriving source from the resolved URL; thread the real `PackageSpec`-derived class through to the cache layer.

- **ruborist**: new `TarballSource { Registry, Http, Git, File }` (`Copy`, no payload) + `From<&PackageSpec>`; `PackageTarballInfo` gains a `source` field (a `CoreVersionManifest` is always `Registry`).
- **install**: classify each lockfile entry by its `resolved` **plus** the set of http-tarball spec URLs collected from every dependency map. An http dep's lockfile `resolved` equals the `https://…` spec its parent declared, while a registry download URL is *never* a declared spec — so the two are distinguished without host heuristics, npm-compatibly.
- **scheduler**: `PackageRef` / `prefetch_download` / `ensure_clone` thread `source`; only `Registry` enters the download pipeline.
- **`resolve_seeded_cache_path`** routes by `source`; a missing http/git/file slot now returns **`Err`** instead of falling through to a registry-slot download — turning silent pollution into a hard "re-run resolution" error.
- retire `is_registry_tarball_url`.

## Tests

- ruborist: `TarballSource::from(&PackageSpec)` variant mapping.
- pm: `tarball_source` classification — a **registry-host tarball URL stays `Registry`** (the exact case the old check got wrong).
- e2e **Case 8.3b**: installs the http-tarball fixture in an isolated cache and asserts each dep lands in `_http_<hash>` and **never** the registry slot `<name>/<version>`.

`cargo fmt` / `clippy -D warnings` (ruborist+pm) / affected unit tests green.

## Open (draft — follow-ups)

- Lockfile-install source recovery relies on the parent spec being present in a dependency map; deeply detached lockfiles fall back to `Registry`. Acceptable since the `Err`-on-missing-slot guard prevents silent pollution regardless.
- `cloner::CacheLayout` still uses `is_git_url`; could move to `TarballSource` too (TODO).

### DIFF
--- a/crates/pm/src/service/install.rs
+++ b/crates/pm/src/service/install.rs
@@ -18,6 +18,7 @@ use crate::model::package::PackageInfo;
 use crate::service::package::PackageService;
 use crate::service::rebuild::RebuildService;
 use crate::util::cli_enum::{OmitType, PackageAction, SaveType};
+use crate::util::downloader::is_git_url;
 use crate::util::json::load_package_lock_json_from_path;
 use crate::util::linker::link;
 use crate::util::logger::{
@@ -25,6 +26,7 @@ use crate::util::logger::{
 };
 use utoo_ruborist::compat::{is_cpu_compatible, is_os_compatible};
 use utoo_ruborist::progress::PackageTarballInfo;
+use utoo_ruborist::spec::{PackageSpec, TarballSource};
 
 use super::binary::update_package_binary;
 use super::clean::clean_deps;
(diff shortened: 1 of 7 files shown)


elrrrrrrr:
Superseded by #3163. This PR fixed the http-tarball cross-cache pollution by isolating remote tarballs in URL-keyed `_http_` cache slots (and threading `TarballSource` to disambiguate registry vs http). #3163 takes the cleaner route discussed offline: non-registry tarballs (untrusted https + local `file:`) are kept out of the global cache entirely — extracted directly into `node_modules` — which removes the same poisoning class plus the path-hashed-slot staleness, and deletes the classification machinery rather than extending it (net −28 lines).

PR #3173, opened 2026-06-17
fix(pm): accept string-form replaceHost in binary-mirror-config

elrrrrrrr:
## Problem

`binary-mirror-config`'s `mirrors.china` map is parsed as **one unit** into a strongly-typed schema (`crates/pm/src/service/binary.rs`). A single entry whose shape drifts from the schema fails the *whole* parse — and `get_envs()` / `update_package_binary()` swallow that failure as a debug log, **silently disabling the entire China mirror layer** (ENVS injection + per-package binary host rewrite) for every package.

The latest upstream config trips exactly this: `flow-bin` ships

```json
"flow-bin": { "replaceHost": "https://github.com/facebook/flow/releases/download/v", "host": "..." }
```

`replaceHost` is a **bare string**, but the field is typed `Option<Vec<String>>`, so deserialization rejects it. Consequence on any non-npm registry (npmmirror / antgroup …): mirror ENVS never get injected and native-binary packages (sharp, node-sass, node-pre-gyp/prebuild consumers) fall back to their original, often-blocked download hosts → install-script failures that surface as "依赖安装失败".

> npm's own registry path is unaffected — it's skipped via `should_skip_binary_mirror()`, which is why plain `binary-mirror-config` installs against npm.org looked fine.

## Fix

- Accept `replaceHost` in **either** shape (single string or array), normalizing to `Vec<String>` — matches cnpm semantics.
- Add a success debug log (`Binary mirror config loaded: N china package entries`) so the layer going dark is observable; previously only the swallowed failure path logged.

## Tests

- Unit: `replaceHost` string/array/absent normalization, plus a full `mirrors.china` parse with the real `flow-bin` string-form entry.
- e2e (bash + pwsh): install `flow-bin` against a non-npm registry with `--verbose`, assert the config loaded and did **not** fail to parse — guards against future upstream schema drift.

### DIFF
--- a/crates/pm/src/service/binary.rs
+++ b/crates/pm/src/service/binary.rs
@@ -49,9 +49,15 @@ struct BinaryMirror {
     #[serde(skip_serializing_if = "Option::is_none")]
     host: Option<String>,
     /// Hosts to rewrite to `host` (used when `replaceHostMap` is absent).
+    /// cnpm ships this in both shapes — a single host string (e.g. `flow-bin`)
+    /// or an array — so accept either and normalize to a list. A strict
+    /// `Vec<String>` would reject the bare-string form and, because the whole
+    /// `mirrors.china` map deserializes as one unit, fail the entire config
+    /// parse — silently disabling mirroring for every package.
     #[serde(
         rename = "replaceHost",
         default,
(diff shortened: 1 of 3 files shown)


killagu (approved):
LGTM

PR #3177, opened 2026-06-17
fix(pm): resolve override targets by protocol, not just registry

elrrrrrrr:
## Problem

`overrides` (and yarn `resolutions`) **target** specs were always resolved as a registry version. Override resolution called `resolve_registry_dep(name, override_spec, …)` directly, bypassing the protocol-aware dispatch that normal dependency specs go through. So any non-registry override target failed:

```jsonc
// overrides: { "is-odd": { "is-number": "file:./is-number.tgz" } }
ERROR Dependency resolution failed: Version resolution failed:
  is-number@file:./is-number.tgz: No matching version found for spec
  'file:./is-number.tgz' from 15 available versions
```

`file:` tarball, http(s) tarball, and git override targets all became bogus registry version lookups → 404 → the whole install aborted with an empty `node_modules`. Only registry ranges / `npm:` aliases / dist-tags worked. This blocked the common local-debugging workflow of `npm pack`-ing a package and pinning it via a (often nested) override.

## Fix

`resolve_override` now parses the override target into a `PackageSpec` and dispatches by protocol, exactly like `process_dependency`:

| target | resolver |
|---|---|
| registry range / `npm:` / dist-tag | registry resolver (**unchanged**) |
| `https://…tgz` | http tarball resolver |
| `git+…` / `github:…` | git resolver |
| `file:…tgz` | parsed into a manifest with a pinned `file:<abs>` tarball, resolved **relative to the root project** (overrides are root-declared), materialized into `node_modules` at install time |
| `file:` **directory** | clear `Unsupported` error (would need a symlink/Link node the override path can't place) — better than the old misleading 404 |

`PackageSpec` only **selects** the resolver; the registry path still receives the raw `override_spec`, so `npm:`/range/dist-tag override behavior is byte-for-byte unchanged. All supported protocols yield a manifest the caller already knows how to place — a tarball manifest just carries a pinned `dist.tarball`, the same shape a normal `file:`/http tarball dep produces.

Optional-edge override failures fall back to the original resolution (npm semantics); load-bearing failures surface the wrapped cause.

## Tests

- Verified locally: nested + direct `file:` tarball overrides land the local build; `file:` dir override gives the clear error; registry override is unregressed.
- 186 ruborist unit tests + 10 doctests pass; clippy clean with **and** without `http-tarball`/`native-git` (wasm gating).
- e2e (bash + pwsh): nested and direct `file:` tarball overrides, asserting the locally-packed version lands and the lockfile pins `file:`.

### DIFF
--- a/crates/ruborist/src/resolver/builder.rs
+++ b/crates/ruborist/src/resolver/builder.rs
@@ -496,6 +496,7 @@ pub async fn process_dependency<R: ManifestProvider>(
                 node_index,
                 edge_info,
                 &resolved.version,
+                config,
                 None,
             )
             .await?
@@ -532,6 +533,7 @@ pub(crate) async fn resolve_override<R: ManifestProvider>(
     parent: NodeIndex,
     edge: &DependencyEdgeInfo,
     resolved_version: &str,
+    config: &BuildDepsConfig,
     mut state: Option<&mut ManifestState>,
 ) -> Result<Option<Arc<CoreVersionManifest>>, ResolveError<R::Error>> {
     let Some(override_spec) = graph.check_override(parent, &edge.name, Some(resolved_version))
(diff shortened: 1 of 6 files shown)


fireairforce (approved):

killagu (approved):
LGTM

PR #3142, opened 2026-06-11
feat(pm): live install progress (download speed, stage gauges, script timing)

elrrrrrrr:
## Summary

Installs previously showed only a package-name spinner — no network feedback, no visibility into long-running downloads/extracts, and lifecycle-script output was discarded on success. This PR adds a live, low-overhead progress line for the whole install pipeline, plus a final traffic/bandwidth summary.

**During install (TTY):**

```
⠼ 86/116 picocolors                         ↓ 2.7 MB 3.5 MB/s · 12 downloading
⠼ 102/116 esbuild postinstall [12s] · 2 more ↓ 23.4 MB 8.2 MB/s · 9 downloading
```

**Final summary:**

```
+ 93 added · 0 reused · 91 downloaded (15.6 MB @ 8.2 MB/s)
```

## Design

- **New `util/install_progress`** — shared, lock-free progress state (zero per-event indicatif contention):
  - process-lifetime `DOWNLOADED_BYTES` counter fed by streaming downloads, with a per-run baseline (`start_install_run` / `downloaded_bytes`) so multi-install processes report only the current run's traffic
  - `DownloadGuard` in-flight download gauge → live `N downloading` concurrency
  - running-script registry (`track_script`) surfacing the *longest-running* script with elapsed time
  - `SpeedMeter` (EMA-smoothed bytes/sec) and pure `compose_activity` / `compose_summary` — both unit-tested
- **Single render task** (`logger`, ~120ms tick) is the only writer of the spinner message; hot paths (downloader chunks, script queue) only touch atomics/registries.
- **Stable layout** — the volatile per-package activity lives in the left `{wide_msg}`; the network summary is pinned to the right edge via `{prefix}` so it doesn't shift as the activity changes. The right segment is a steady throughput readout (`↓ <cumulative> <speed> · N downloading`) — deliberately not the extract/link in-flight gauges, which blink on/off as batches drain.
- **Streaming downloads** — `downloader` switches from `response.bytes()` to `bytes_stream()` (capacity hint capped at 32 MB against a bogus Content-Length), feeding the byte counter as chunks arrive.
- **Script output auditability** — successful lifecycle-script stdout/stderr now lands in the debug log (`$TMPDIR/utoo-<ts>.log`), capped at 64 KiB per stream.
- **Console logs no longer garble the bar** — the tracing console layer flushes each event inside `PROGRESS_BAR.suspend()`. A succeeded download retry now logs at `debug` (not `info`) so it doesn't interrupt the spinner.
[...]

PR #3178, opened 2026-06-17
fix(pm): cache binary-mirror-config load and accept string replaceHost

elrrrrrrr:
## Problem

On any **non-npmjs registry** (npmmirror / cnpm / aliyun mirrors — exactly where binary-mirror is meant to work), `utoo install`/`update` crawls and appears to hang. Surfaced while accepting #3142 on a large tree (ant-design): the spinner sat on the clone phase (`2720/5111 <pkg> resolved`) for ~92s. The debug log showed **838 repeated** `Binary mirror config unavailable` lines on the main thread, ~100ms apart.

Root cause is two compounding regressions from **#3157** (`service/binary.rs`):

1. **Uncached failure → per-package re-fetch (the stall).** `load_config` used `OnceCell::get_or_try_init`, which does **not** memoize an `Err`. Since `update_package_binary` runs once per cloned package, a parse failure re-ran the closure — a fresh network fetch of `binary-mirror-config@latest` — for *every* package, serialized on the main thread.
2. **Strict type rejects npm's polymorphic config (the parse failure).** `BinaryMirror.replace_host` was typed `Option<Vec<String>>`, but npm's `binary-mirror-config` has `flow-bin.replaceHost` as a **bare string** (`replaceHost` is string-or-array). The strict type failed the whole parse, silently disabling china-mirror rewrites.

Pre-#3157 the config was parsed as untyped `serde_json::Value`, which never failed — so this is a regression introduced by the strong-typing refactor.

## Fix

- Store `OnceCell<Result<BinaryMirrorConfig, String>>` and use `get_or_init`, so a fetch/parse failure is cached — **at most one attempt per process**.
- Add a `de_string_or_seq` deserializer so `replaceHost`/`replaceHostFiles` accept a string *or* an array, normalizing to `Vec<String>`.

## Tests

- Unit: `test_replace_host_accepts_string_or_array` (field-level) and `test_full_config_with_string_replace_host_parses` (whole-config regression).
- **e2e:** new `e2e/pm/binary-mirror` case installs `utf-8-validate` against `registry.npmmirror.com` and asserts the config parses — zero `Binary mirror config unavailable` lines in the run's debug log.

Verified locally against the real npmmirror config: **838 redundant fetches → 0**, install of the fixture in ~2s.

> Independent of #3142 (live install progress); that PR only *displayed* this stall faithfully.

### DIFF
--- a/crates/pm/src/service/binary.rs
+++ b/crates/pm/src/service/binary.rs
@@ -4,10 +4,11 @@ use crate::util::json::read_json_file;
 use crate::util::user_config::get_registry;
 use anyhow::{Context, Result};
 use regex::Regex;
-use serde::Deserialize;
+use serde::{Deserialize, Deserializer};
 use serde_json::{Map, Value};
 use std::collections::BTreeMap;
 use std::path::Path;
+use std::result::Result as StdResult;
 use std::sync::OnceLock;
 use tokio::sync::OnceCell;
 use utoo_ruborist::registry::is_npm_registry;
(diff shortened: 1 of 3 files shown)


elrrrrrrr:
Superseded by #3173, which already merged the same two fixes to `next` (string-or-array `replaceHost` + caching `load_config` failures via `get_or_init`). #3173's e2e coverage is broader (bash + pwsh, installs flow-bin directly), so this PR adds nothing. Closing as duplicate.

PR #3180, opened 2026-06-18
fix(pm): normalize registry URL at resolution chokepoint

elrrrrrrr:
## Problem

Registry values from the CLI (`--registry`), `UTOO_REGISTRY`, and the config file arrive unnormalized. A trailing slash is common — npm's own default is `https://registry.npmjs.org/`, and `utooland/setup-utoo` passes exactly that into `ut config set registry`. Downstream code joins package paths as `{registry}/{pkg}`, so a trailing slash produces a double slash:

```
https://registry.npmjs.org//@svgr/babel-plugin-remove-jsx-empty-expression
```

npm's registry rejects the `//<pkg>` path with **HTTP 406 Not Acceptable**, aborting dependency resolution. Observed in ant-design CI:

```
ERROR Dependency resolution failed: Registry error: Failed to fetch
@svgr/babel-plugin-remove-jsx-empty-expression: HTTP 406 Not Acceptable:
https://registry.npmjs.org//@svgr/babel-plugin-remove-jsx-empty-expression
```

## Fix

Canonicalize the registry **once**, in `init_registry` — the single point where all sources (CLI > env > config > auto-select) converge:

- `Url::parse` to validate, rejecting non-`http(s)`/unparsable values early with a clear message instead of a late, opaque fetch failure;
- strip trailing slashes so `{registry}/{pkg}` never doubles up;
- **preserve sub-paths**, so private registries served under a prefix (Artifactory/Nexus/Verdaccio `…/api/npm/<repo>`) keep working;
- fail fast when an explicitly configured registry is invalid, rather than silently falling back to the public default registry.

Reuses pm's existing `reqwest::Url` (no new dependency). Adds unit tests for trailing-slash stripping, sub-path preservation, and invalid-input rejection.

## Notes

The upstream trigger was also fixed in `utooland/setup-utoo` (default registry no longer carries a trailing slash); this PR makes `ut` robust regardless of what callers pass.


elrrrrrrr:
@codex review

elrrrrrrr:
@codex review

killagu (approved):
LGTM

fireairforce (approved):

PR #3182, opened 2026-06-18
perf(pm): shrink binary 10MB→6.4MB — system git, single TLS provider, drop hickory/h2

elrrrrrrr:
> **Draft / investigation.** All three trims are implemented, tested, and benched. Opening as draft for review of the tradeoffs (notably: `git` is now a runtime requirement for git deps).

## Result

| build (macOS arm64, `release` profile) | size |
|---|---|
| baseline (`next`) | 10 MB |
| + single TLS provider + drop hickory | 8.6 MB |
| **+ system git (drop gix + h2)** | **6.4 MB** |

**−3.6 MB / −36%.** The new `pm-binary-size` GHA job reports Linux + macOS sizes and asserts the trimmed dep set on both.

## 1. Git deps via system `git` (biggest win, ~2.2 MB)

Git dependencies were cloned by bundled **`gix`** (a full Rust git impl). Switched to shelling out to the user's `git` — exactly what **npm/pnpm/yarn do**. gix dragged in a ~20-crate family, *plus* `gix-transport`'s default-featured reqwest (the sole non-dev puller of **`h2`/HTTP-2**), *plus* `bstr`→regex-automata. All gone.

- `resolver/git.rs` reimplemented on `std::process::Command("git")`: shallow `clone --depth 1 --branch <ref>` for branch/tag/HEAD; full clone + `checkout` for a SHA. `rev-parse HEAD` pins `<cache>/<name>/<sha>/`; working tree copied (exec bits + symlinks preserved, `.git` skipped, path-traversal guard kept). **Same public API + cache layout** — no façade/pm changes.
- **Tradeoff:** `git` is now required for `git:`/`github:` deps; a missing binary errors with a clear message naming it (npm is the same). Registry installs need no git.
- Verified: e2e git-deps resolves all three spec forms (`github:owner/repo`, `git+https#tag`, `owner/repo#tag`) to the correct pinned commit.

## 2. One TLS provider — `ring` (~0.6 MB)

macOS was compiling **two** rustls crypto providers: `aws-lc-rs` (forced by ruborist) **and** `ring` (pulled by pm's `rustls-tls`). Dropped the macOS `aws-lc-rs` override → all native targets use `ring`.

The `aws-lc-rs` special-case existed for "~40% faster M-series cold resolves" — **that didn't reproduce.** Same-machine A/B (ant-design-sized cold resolve, cache cleared, 5 interleaved runs):

| registry | ring (median/min) | aws-lc-rs (median/min) | aws-lc vs ring |
|---|---|---|---|
| registry.npmmirror.com | **1.47s / 1.32s** | 1.87s / 1.83s | **+27% slower** |
| registry.antgroup-inc.cn | 2.59s / 2.26s | 2.84s / 2.19s | +10% median / tied min (noisy) |

cold resolve is network-bound, so aws-lc's crypto edge is invisible while its per-client `load_native_certs` + `install_default` adds fixed overhead. So this is a size win **and** a small perf win on China registries.

## 3. Drop unused `hickory-dns` (~0.5 MB)

Both clients already set `.dns_resolver(shared_resolver())` (custom getaddrinfo + 300s cache), so reqwest's `hickory-dns` feature was compiled but **never used** — pure dead weight, zero behavior change, helps every platform.

## Research — lighter reqwest / tokio?

**Not viable.** reqwest is heavy but load-bearing for the resolver's async concurrent-fetch model, and reqwest 0.12 *requires* tokio; both are deeply embedded. Sync HTTP (ureq) would abandon the concurrency and still need rustls. (Note: with gix gone, reqwest is now pulled only by pm/ruborist, so it *could* in principle be swapped — but the cost/risk still isn't worth it.) The realistic levers were the dependency trims in this PR.

## Verification

- Local macOS arm64 `release` → **6.4 MB**; `cargo tree -e no-dev` confirms `gix`/`gix-transport`/`h2`/`aws-lc-sys`/`hickory-*` all absent, `ring` sole provider.
- A/B cold-resolve bench (above) on both registries.
- e2e git-deps (3 spec forms) resolve via system git; `cargo clippy --all-targets -D warnings` clean; `cargo test -p utoo-pm` 272 passed + 13 ruborist git tests.
- `pm-binary-size` GHA workflow (ubuntu + macos) reports size + asserts the trimmed set.

### DIFF
--- a/.github/workflows/pm-binary-size.yml
+++ b/.github/workflows/pm-binary-size.yml
@@ -1,0 +1,69 @@
+name: utoopm-binary-size
+
+# Reports the release `utoo` binary size on Linux + macOS and asserts that the
+# trimmed dependency set (no aws-lc-rs, no hickory) holds. Lets a size-focused
+# PR show the real shipped-artifact size on both platforms, since local builds
+# are usually one arch only. Manual + on PRs that touch the relevant crates.
+on:
+  workflow_dispatch:
+  pull_request:
+    paths:
+      - "crates/pm/**"
+      - "crates/ruborist/**"
+      - "Cargo.toml"
+      - "Cargo.lock"
+      - ".github/workflows/pm-binary-size.yml"
+
+jobs:
+  size:
+    name: ${{ matrix.target }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - os: ubuntu-latest
+            target: x86_64-unknown-linux-gnu
+          - os: macos-latest
+            target: aarch64-apple-darwin
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install Rust
+        uses: dtolnay/rust-toolchain@stable
+        with:
+          toolchain: nightly-2026-04-02
+          targets: ${{ matrix.target }}
+
+      - name: Init git submodules
+        run: git submodule update --init --recursive --depth 1
+
+      - name: Build release binary
+        run: cargo build --release --bin utoo --target ${{ matrix.target }} -p utoo-pm
+
+      - name: Report size
+        run: |
+          BIN="target/${{ matrix.target }}/release/utoo"
+          echo "## utoo release binary — ${{ matrix.target }}" >> "$GITHUB_STEP_SUMMARY"
+          ls -lh "$BIN"
+          BYTES=$(wc -c < "$BIN")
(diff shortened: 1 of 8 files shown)


elrrrrrrr:
Superseded by #3183 (codex), which takes the same approach (system git, ring TLS, drop hickory/http2) but is a superset: it also trims `tokio/full` → concrete features and reqwest `stream` (slightly smaller, 6.32 MB vs 6.4 MB), hardens the git fallback (SHA-256 + `--`), and folds the size-assert into pm-ci.yml. #3183's CI is green; this branch accidentally bumped the next.js submodule pointer. Posted my ring-vs-aws-lc A/B bench data to #3183 as the empirical justification for dropping aws-lc. Closing in favor of #3183.

PR #3183, opened 2026-06-18
perf(pm): reduce utoo-pm binary size

elrrrrrrr:
## Summary

* Remove the gix/gitoxide HTTP transport from `utoo-ruborist` native git resolution by delegating clone/checkout to the system `git` executable, while keeping the existing cache layout, token injection, dedup cache, manifest finalization, and atomic cache commit protocol.
* Trim `utoo-pm`/`utoo-ruborist` release features: drop reqwest `http2`, `hickory-dns`, and `stream`; remove the macOS aws-lc TLS override; keep socks proxy support; replace `tokio/full` with the features the PM code actually uses; align PM on workspace `dirs`.
* Add a `utoopm-binary-size` GitHub Actions job that asserts `h2`, `hickory-resolver`, `gix`, and `aws-lc-rs` stay out of the Linux release dependency tree and reports the release binary size.
* Address git fallback hardening by passing `--` before clone positional args and accepting both 40-char SHA-1 and 64-char SHA-256 full hashes.

## Local Size Result

Measured in this worktree with `cargo build -p utoo-pm --release --bin utoo` on macOS:

* Baseline from `origin/next`: `10,419,376` bytes
* Optimized: `6,319,232` bytes
* Delta: `-4,100,144` bytes (`-39.35%`)

The local Linux-target release-tree assertion also passes:

```bash
cargo tree -p utoo-pm --target x86_64-unknown-linux-gnu -e normal,build --prefix none
# no h2 / hickory-resolver / gix / aws-lc-rs entries
```

## Reqwest / Tokio Alternatives Notes

* `reqwest` is still the lowest-risk fit for the async registry clients because the code relies on async requests, shared client builders, proxy handling, socks proxy support, gzip/json, rustls, and custom DNS resolver policy. The important win here is feature diet rather than replacing it. Reqwest docs also call out socks as an explicit feature for socks proxy envs, so this PR keeps it: https://docs.rs/reqwest/
* `ureq` is a plausible future experiment for isolated blocking download paths, but it is blocking by design and would need `spawn_blocking` plus new pooling/proxy/retry integration for the resolver: https://docs.rs/ureq
* `minreq` is even smaller/minimal, but too bare for the current registry client behavior without rebuilding much of the HTTP policy stack: https://docs.rs/minreq
* `tokio` should stay; PM uses process, fs, net, sync, time, select, and the multi-thread runtime. The size win is replacing `full` with a concrete feature list. Tokio's feature list shows `full` includes extras such as `io-std`, `parking_lot`, and `signal` that PM release code does not need: https://docs.rs/crate/tokio/latest/features

## Validation

* `cargo fmt`
* `tombi format Cargo.toml crates/pm/Cargo.toml crates/ruborist/Cargo.toml`
* `cargo clippy --all-targets -- -D warnings --no-deps`
* `cargo test -p utoo-ruborist --features native-git -- --nocapture` (`201 passed` before the follow-up hardening)
* `cargo test -p utoo-ruborist --features native-git full_sha_detection_accepts_sha1_and_sha256 -- --nocapture`
* `cargo test -p utoo-ruborist --features native-git ensure_repo_cached_clones_local_repo -- --nocapture`
* `cargo test -p utoo-pm -- --nocapture` (`272 passed`, `3 ignored`)
* `cargo build -p utoo-pm --release --bin utoo`
* Local copy of the new GHA release dependency assertion

### DIFF
--- a/.github/workflows/pm-ci.yml
+++ b/.github/workflows/pm-ci.yml
@@ -14,6 +14,51 @@ permissions:
   contents: read
 
 jobs:
+  binary-size:
+    if: (github.event_name == 'push' && contains(github.event.head_commit.message, '(pm)')) || (github.event_name == 'pull_request' && contains(github.event.pull_request.title, '(pm)'))
+    name: utoopm-binary-size
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Init git submodules
+        run: git submodule update --init --recursive --depth 1
+
+      - name: Install Rust
+        uses: dtolnay/rust-toolchain@stable
+        with:
+          toolchain: nightly-2026-04-02
+          targets: x86_64-unknown-linux-gnu
+
+      - name: Assert release dependency trim
+        shell: bash
+        run: |
+          set -euo pipefail
+          mkdir -p target
+          cargo tree -p utoo-pm --target x86_64-unknown-linux-gnu -e normal,build --prefix none > target/utoo-pm-release-tree.txt
+
+          if grep -E '^(h2|hickory-resolver|gix|aws-lc-rs) ' target/utoo-pm-release-tree.txt; then
+            echo "Unexpected heavyweight dependency found in utoo-pm release tree" >&2
+            exit 1
+          fi
+
+      - name: Build release binary
+        run: cargo build -p utoo-pm --release --bin utoo --target x86_64-unknown-linux-gnu
+
+      - name: Report binary size
+        shell: bash
+        run: |
+          set -euo pipefail
+          binary="target/x86_64-unknown-linux-gnu/release/utoo"
+          bytes="$(stat -c '%s' "$binary")"
+          {
+            echo "### utoo-pm binary size"
+            echo
+            echo "- target: x86_64-unknown-linux-gnu"
+            echo "- bytes: $bytes"
+            echo "- binary: $binary"
+          } >> "$GITHUB_STEP_SUMMARY"
+
   test:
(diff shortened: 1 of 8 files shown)


elrrrrrrr:
@codex review

elrrrrrrr:
### Empirical backing for dropping the macOS `aws-lc-rs` override

The `aws-lc-rs` special-case existed for a "~40% faster M-series cold resolve" claim. I A/B-tested that on a separate branch before this PR existed — sharing the data here since it justifies the TLS change with numbers (the perf claim **did not reproduce**).

**Method:** same machine (macOS arm64, M-series), two release binaries differing only in the rustls provider at runtime (ring vs aws-lc-rs — DNS/h2 unused at runtime in both). Cold resolve of an ant-design-sized tree (~659 pkgs, caret ranges), manifest cache cleared each run, 5 interleaved runs.

| registry | ring (median / min) | aws-lc-rs (median / min) | aws-lc vs ring |
|---|---|---|---|
| registry.npmmirror.com | **1.47s / 1.32s** | 1.87s / 1.83s | **+27% slower** |
| registry.antgroup-inc.cn | 2.59s / 2.26s | 2.84s / 2.19s | +10% median / tied at min (noisy) |

cold resolve is network-bound, so aws-lc's crypto edge is invisible, while its per-client `load_native_certs` + `install_default` adds fixed overhead. So unifying to `ring` is a size win **and** a small perf win on China registries — no tradeoff on the registries that matter here. (The original ~40% likely came from npmjs.org / a different setup.)

Superseding my #3182 in favor of this PR — it's a superset (also trims `tokio/full` → concrete features + `stream`, plus SHA-256 / `--` clone hardening). Closing #3182.

killagu (approved):
LGTM

fireairforce (approved):

PR #3186, opened 2026-06-21
fix(pack): resolve CJS require() to main field on node target, not the ESM module field

### LINKED ISSUE #3185: 🐛 CJS require() of an ESM-only-default dependency returns the namespace under node target (json5 `JSON5.parse is not a function`)
## What happened?

When bundling for **node** (`target: node 22`, `platform: node`, `output.type: standalone`), an **authored-CJS** module that does `require('pkg')` on a dependency whose **only ESM export is `default`** receives the ESM **namespace object** (`{ default, __esModule }`) instead of the default's members. Named access is therefore `undefined`.

This is the real-world `<path>/tsconfig.json is malformed JSON5.parse is not a function` crash when bundling [cnpmcore](https://github.com/cnpm/cnpmcore) (an egg app) with `@utoo/pack`:

- `tsconfig-paths/lib/tsconfig-loader.js` does `const JSON5 = require('json5'); JSON5.parse(...)`.
- `json5` ships `main: lib/index.js` (CJS, `module.exports = { parse, stringify }`), `module: dist/index.mjs` (ESM, **only** `export default`), and **no `exports` field**.
- `@utoo/pack` resolves the `module` (ESM) field for the `require()`, so the CJS caller gets `{ default: { parse } }` and `JSON5.parse` is `undefined`.

It is a generic *"CJS `require()` of an ESM-only-`default` dependency"* interop defect, json5 is just the dependency that happens to trip it.

### Environment

- `@utoo/pack` 1.4.13 (also reproduces conceptually on 1.4.14)
- `target: node 22`, `platform: node`, `output.type: standalone`

### Minimal reproduction

A dependency shaped exactly like `json5`:

```jsonc
// node_modules/esm-default-only/package.json
{ "name": "esm-default-only", "main": "index.js", "module": "index.mjs" }  // NO "exports"
```

```js
// index.js   (CJS main)
module.exports = { parse: (s) => 'cjs:' + s };
// index.mjs  (ESM module — ONLY a default export)
export default { parse: (s) => 'esm:' + s };
```

An authored-CJS consumer, exactly like tsconfig-paths uses json5:

```js
// consumer.js
const JSON5 = require('esm-default-only');
module.exports = () => typeof JSON5.parse; // expects 'function'
```

Build entry → consumer with `target: node 22`, `platform: node`, `output.type: standalone`, then run the output.

A standalone runnable script (drives a real `@utoo/pack` build via `cjs/commands/build.js`, scaffolds the fixture, executes the output, asserts) is attached below.

<details>
<summary>repro.mjs</summary>

```js
import fs from 'node:fs';
import path from 'node:path';
import { fileURLToPath } from 'node:url';
import { createRequire } from 'node:module';
import { execFileSync } from 'node:child_process';

const __dirname = path.dirname(fileURLToPath(import.meta.url));
const require = createRequire(import.meta.url);

// Point this at your @utoo/pack build command (CJS entry).
const PACK_BUILD = process.argv[2] || require.resolve('@utoo/pack/cjs/commands/build.js');
const { build: packBuild } = require(PACK_BUILD);

const ROOT = path.join(__dirname, '.work');
fs.rmSync(ROOT, { recursive: true, force: true });
const SRC = path.join(ROOT, 'src');
const OUT = path.join(ROOT, 'dist');
const DEP = path.join(ROOT, 'node_modules', 'esm-default-only');
fs.mkdirSync(SRC, { recursive: true });
fs.mkdirSync(DEP, { recursive: true });
const w = (p, c) => fs.writeFileSync(p, c);

w(path.join(DEP, 'package.json'), JSON.stringify({ name: 'esm-default-only', version: '1.0.0', main: 'index.js', module: 'index.mjs' }));
w(path.join(DEP, 'index.js'), "module.exports = { parse: (s) => 'cjs:' + s };\n");
w(path.join(DEP, 'index.mjs'), "export default { parse: (s) => 'esm:' + s };\n");
w(path.join(SRC, 'consumer.js'), "const JSON5 = require('esm-default-only');\nmodule.exports = function run() { return { type: typeof JSON5.parse, value: typeof JSON5.parse === 'function' ? JSON5.parse('ok') : null }; };\n");
w(path.join(SRC, 'entry.js'), "const run = require('./consumer');\nconsole.log('REPRO_RESULT ' + JSON.stringify(run()));\n");

fs.mkdirSync(OUT, { recursive: true });
w(path.join(OUT, 'package.json'), JSON.stringify({ type: 'commonjs' }));

const config = {
  entry: [{ name: 'worker', import: path.join(SRC, 'entry.js') }],
  target: 'node 22',
  platform: 'node',
  mode: 'production',
  output: { path: OUT, type: 'standalone', clean: false },
  optimization: { moduleIds: 'named', minify: false },
  sourceMaps: false,
};

await packBuild({ config }, ROOT, ROOT);

const out = execFileSync('node', [path.join(OUT, 'worker.js')], { encoding: 'utf8' });
const result = JSON.parse(out.match(/REPRO_RESULT (\{.*\})/)[1]);
console.log(result.type === 'function'
  ? `PASS  named access works (value=${JSON.stringify(result.value)})`
  : `FAIL  named access is '${result.type}' — JSON5.parse is not a function`);
process.exit(result.type === 'function' ? 0 : 1);
```

</details>

### Observed output (1.4.13)

```
bundled esm entry? : true   (index.mjs / module field)
bundled cjs entry? : false  (index.js / main field)
runtime result     : { "type": "undefined", "value": null }
FAIL  named access is 'undefined' — JSON5.parse is not a function
```

The emitted chunk shows the cause — `commonJsRequire` (`.r`) returns the ESM module's `module.exports`, which is the namespace object:

```js
// node_modules/esm-default-only/index.mjs -> ESM module, namespace { default }
var __TURBOPACK__default__export__ = { parse: (s) => 'esm:' + s };
__turbopack_context__.s(["default", () => __TURBOPACK__default__export__]);

// consumer.js
const JSON5 = __turbopack_context__.r(".../index.mjs ... (ecmascript)");
//            ^ resolves to { default: { parse } }  =>  JSON5.parse === undefined
```

## Expected result

A CJS `require()` of such a dependency should yield a usable value, i.e. `JSON5.parse` is a function — matching Node's own behavior for these packages (Node resolves the CJS `main` for a `require()`).

## Possible fix angles

1. **Resolver (root cause).** For a `require()` (CJS) context under a node target, prefer the package's `main` / `require` condition over `module` / `import`, so json5 resolves to its CJS entry and no interop is needed. Today `ResolveOptions` only exposes `alias` + `extensions` (no `mainFields` / `conditionNames`), so consumers cannot opt into this from config.

2. **Runtime interop.** When `commonJsRequire` resolves to a **genuine ESM module** whose only own export is `default`, return `.default`. Gate on `module.namespaceObject === module.exports` so transpiled-CJS modules (`Object.defineProperty(exports,'__esModule',…); exports.default = X`, where the consumer *does* want `.default`) are left untouched — only real ESM modules set `namespaceObject`.

For context: the egg side currently works around this with an output post-pass that rewrites `commonJsRequire` using angle #2, but fixing it in `@utoo/pack` (either angle) would remove the need for that and benefit all node-target consumers.

killagu:
## Problem

Fixes #3185.

When bundling for a **node** target, an authored-CJS module that does `require('pkg')` on a dependency whose only ESM export is `default` (and which has **no `exports` field**) received the ESM **namespace object** (`{ default, __esModule }`) instead of the CJS `main`'s value. Named access was therefore `undefined`.

Real-world crash: bundling [cnpmcore](https://github.com/cnpm/cnpmcore) → `tsconfig-paths` does `const JSON5 = require('json5'); JSON5.parse(...)`. `json5` ships `main: lib/index.js` (CJS), `module: dist/index.mjs` (ESM, only `export default`), and **no `exports`**. Pack resolved the `module` (ESM) field for the `require()`, so the caller got `{ default: { parse } }` and `JSON5.parse` was `undefined` → `JSON5.parse is not a function`.

## Root cause

`get_server_resolve_options_context` set `module: true`. In turbopack's resolver (`resolve.rs`), the `into_package` order is `exports` → `browser` → **`module`** → `main`, and `apply_cjs_specific_options` only flips the `import`/`require` conditions **inside the `exports`/`imports` condition maps** — it does **not** remove the `module` main-field entry for a `require()`. So for a package without `exports`, both `require()` and `import` resolved the ESM `module` field first.

**Node honors neither the `module` main field nor the `module` exports condition** — both are bundler-only conventions. Enabling them on a node target diverges from Node and produces this interop defect (it also breaks `import { parse } from 'json5'`, which would get `undefined` for the same reason).

## Fix

Disable the `module` field on the server (node) resolve context (`module: false`). Resolution now matches Node in both `require()` and `import` directions. Packages that ship an `exports` field are unaffected — `exports` takes priority over main fields.

## Verification

Self-contained repro from the issue (CJS `require()` of a json5-shaped dep, node-target standalone build, then run the output):

| Build | Result |
|---|---|
| Published `@utoo/pack@1.4.14` | ❌ `FAIL  named access is 'undefined'` |
| This branch (local build) | ✅ `PASS  named access works (value="cjs:ok")` |

`value="cjs:ok"` confirms the `require()` now resolves the CJS `main` (which prepends `cjs:`) rather than the ESM `module` build.

Also verified locally:
- `cargo fmt --check` — clean
- `cargo clippy -p pack-core --all-targets -- -D warnings --no-deps` — clean

## CI dependency

CI's `clippy` job is currently red on `next` due to a **pre-existing, unrelated** break (a9ee35bb added `bytes_stream()` without reqwest's `stream` feature). That is fixed separately in **#3187**. Once #3187 lands, this branch will be rebased onto `next` and CI will be fully green.

### DIFF
--- a/crates/pack-core/src/server/contexts.rs
+++ b/crates/pack-core/src/server/contexts.rs
@@ -352,7 +352,13 @@ pub async fn get_server_resolve_options_context(
         custom_conditions,
         import_map: Some(server_import_map),
         fallback_import_map: Some(server_fallback_import_map),
-        module: true,
+        // Node honors neither the `module` main field nor the `module` exports condition; both are
+        // bundler-only conventions. Enabling them on a node target makes a `require()` (or `import`)
+        // of a package without an `exports` field resolve the ESM `module` build instead of the CJS
+        // `main` Node would load — yielding the ESM namespace (`{ default, __esModule }`) so named
+        // access becomes `undefined` (e.g. `require('json5').parse`). Keep it off to match Node.
+        // See https://github.com/utooland/utoo/issues/3185
+        module: false,
         before_resolve_plugins: vec![ResolvedVc::upcast(externals_plugin)],
         after_resolve_plugins: vec![ResolvedVc::upcast(externals_plugin)],
         ..Default::default()


fireairforce (approved):
lgtm

PR #3188, opened 2026-06-21
perf(pm): drop reqwest `stream` feature, stream download via `Response::chunk`

elrrrrrrr:
## Background

`crates/pm/src/util/downloader.rs` reported live install progress via `response.bytes_stream()` (added in #3142 / a9ee35bb), which requires reqwest's `stream` feature. #3183 (binary slimming) had removed `stream` on a base that predated that call site, so once both PRs landed on `next` the combination failed to compile:

```
error[E0599]: no method named `bytes_stream` found for struct `reqwest::Response`
```

#3187 unblocked `next` by **re-adding** the `stream` feature. This PR does it properly instead.

## Why progress never needed `stream`

`reqwest::Response::chunk()` reads the body frame-by-frame with **no feature gate** (only `bytes_stream()` is `#[cfg(feature = "stream")]`). It feeds the exact same per-chunk `DOWNLOADED_BYTES` counter the spinner renders as `↓ 23.4 MB 8.2 MB/s`.

The body is still **fully buffered** before return — the downstream extractor is not streaming (the decompression path isn't incremental) — so `chunk()` only buys the progress signal, not lower peak memory. Behavior is byte-for-byte identical: npm `.tgz` aren't transfer-`gzip`-encoded, so both paths count raw tarball bytes the same.

## Changes

- Revert #3187's `stream` feature (the `Cargo.lock` diff is the exact inverse of #3187 — only the wasm-only `wasm-streams` dep is dropped; native binary size is unaffected).
- Rewrite the download loop to use `Response::chunk()`.
- Drop the now-unused `futures::StreamExt` import.

## Verification

```
cargo check  -p utoo-pm                                   # ok
cargo clippy -p utoo-pm --all-targets -- -D warnings --no-deps   # ok
cargo fmt                                                 # no changes
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6140,7 +6140,6 @@ dependencies = [
  "url",
  "wasm-bindgen",
  "wasm-bindgen-futures",
- "wasm-streams",
  "web-sys",
 ]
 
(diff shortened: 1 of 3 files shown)


killagu (approved):
LGTM

PR #3217, opened 2026-07-06
feat(pm): read registry configuration from .npmrc

### LINKED ISSUE #3212: feat(pm): support reading registry configuration from .npmrc
## Problem

utoo does not read `.npmrc` at all, so a project-level `registry=` setting is silently ignored and installs resolve against the default public registry.

Demo:

```bash
cd my-app
cat .npmrc
# registry=https://registry.npmmirror.com

ut install
# .npmrc is ignored: utoo resolves against its own configured/auto-selected registry
```

Registry resolution (`init_registry` in `crates/pm/src/util/user_config.rs`) only considers:

1. `--registry` CLI flag
2. `UTOO_REGISTRY` env
3. utoo's own config (`~/.utoo/config.toml`, `.utoo.toml`)
4. auto-selected fastest public registry

There is no `.npmrc` parsing anywhere in the codebase (the only mention is in `pm_pack.rs`, which excludes the file from published tarballs). Auth tokens are likewise only read from `NPM_TOKEN` / `NODE_AUTH_TOKEN` env vars or utoo config keys, never from `//host/:_authToken=`.

## Real-world failure

vite-plus ecosystem-ci publishes preview builds to a dedicated preview registry and auto-adds an `.npmrc` with `registry=` pointing at it. Repos installing with `ut` (via `utooland/setup-utoo`) fail because utoo ignores the `.npmrc` and queries the public registry, where the preview version does not exist:

```
ERROR Dependency resolution failed: Registry error: vite-plus@0.0.0-commit.1914fa81…:
No matching version found for spec '0.0.0-commit.1914fa81…' from 136 available versions
```

Example run: https://github.com/vite-plus-ecosystem-ci/cnpmcore/actions/runs/28726133745/job/85183773987 (from https://github.com/vite-plus-ecosystem-ci/cnpmcore/pull/3)

Any npm-compatible client is generally expected to honor `.npmrc`, so ecosystem tooling (registry bridges, private registries, corporate proxies) assumes it works.

## Proposal

Teach utoo to read `.npmrc` as a fallback layer in registry resolution, following npm's file precedence (project > user `~/.npmrc` > global), slotted between `UTOO_REGISTRY` and utoo's own config (or after utoo config, to be discussed):

- `registry=` (minimum viable fix, unblocks ecosystem-ci)
- `@scope:registry=` scoped registries
- `//host/:_authToken=` credentials

## Workaround

Set `UTOO_REGISTRY` (e.g. `UTOO_REGISTRY=https://registry.npmmirror.com`) in the environment, or write the registry into `.utoo.toml`.

fengmk2:
Closes #3212

utoo now reads npm's `.npmrc` files, so registry bridges, private registries, and corporate proxies work with `ut` without utoo-specific setup. Supported subset:

- `registry=` from the project (`<cwd>/.npmrc`) and user (`~/.npmrc`) files
- `//host[/path]/:_authToken=` credentials, matched with npm's nerf-dart walk-up rules (most specific path wins, scheme-less registry values supported)
- `${VAR}` environment interpolation in values (an unset var drops that entry with a warning)

Registry resolution precedence, interleaved by level so a project `.npmrc` can redirect any npm client while utoo-native config stays authoritative at the same level:

1. `--registry` CLI flag
2. `UTOO_REGISTRY` env
3. project `.utoo.toml`
4. project `.npmrc`
5. global `~/.utoo/config.toml`
6. user `~/.npmrc`
7. auto-select fastest registry

Auth tokens resolve `NPM_TOKEN`/`NODE_AUTH_TOKEN` > `ut login` config > `.npmrc`, and flow to both manifest fetches and tarball downloads through the existing plumbing.

Implementation notes: the parsing lives in a new pm-only `util/npmrc.rs` module (wasm build untouched), and `Config::load_levels()` now reads each utoo config file once, populating the merged cache and returning per-level views for the interleaving. `@scope:registry=` routing is deferred to a follow-up, since resolution currently uses a single registry per install (`UnifiedRegistry` holds one URL).

Tested with unit tests for parsing, token matching, and precedence, plus a new e2e case: a stub registry that rejects any request without `Authorization: Bearer <token>`, where the registry URL and token come only from the project `.npmrc` (token via `${E2E_NPMRC_TOKEN}`), proving the override, the auth flow, and env interpolation end-to-end. This fixes the vite-plus ecosystem-ci failure mode where a bridge `.npmrc` was silently ignored.

### DIFF
--- a/crates/pm/src/service/auth.rs
+++ b/crates/pm/src/service/auth.rs
@@ -5,6 +5,7 @@ use tokio::sync::OnceCell;
 use crate::util::cli_enum::ConfigScope;
 use crate::util::config_file::Config;
 use crate::util::http::client;
+use crate::util::npmrc;
 
 fn registry_api(registry: &str, path: &str) -> String {
     format!("{}{}", registry.trim_end_matches('/'), path)
@@ -34,6 +35,10 @@ fn token_key(registry: &str) -> String {
 }
 
 /// Resolve auth token for a specific registry.
+///
+/// Priority: `NPM_TOKEN`/`NODE_AUTH_TOKEN` env > utoo config (`ut login`
+/// writes `auth-token:<host>` to the global config) > `.npmrc`
+/// `//host/:_authToken=` credentials (project file first, npm walk-up rules).
 pub async fn resolve_token(registry: &str) -> Option<String> {
     for var in ["NPM_TOKEN", "NODE_AUTH_TOKEN"] {
         if let Ok(token) = std::env::var(var)
@@ -43,9 +48,14 @@ pub async fn resolve_token(registry: &str) -> Option<String> {
         }
     }
 
-    let config = Config::load(ConfigScope::Global).await.ok()?;
-    let token = config.get(&token_key(registry)).ok().flatten()?;
-    if token.is_empty() { None } else { Some(token) }
+    if let Ok(config) = Config::load(ConfigScope::Global).await
+        && let Ok(Some(token)) = config.get(&token_key(registry))
+        && !token.is_empty()
+    {
+        return Some(token);
+    }
+
+    npmrc::load().await.auth_token_for(registry)
 }
 
 /// Token for the configured registry, resolved once and cached for the process.
@@ -75,14 +85,21 @@ pub(crate) fn parse_registry(registry: &str) -> Option<Url> {
     }
 }
 
-/// Whether `url` targets the same host as `registry`. The leak guard for
-/// attaching a token to a tarball download: credentials go to the registry
-/// host only, never to a third-party CDN a manifest's `dist.tarball` points at.
+/// Whether `url` targets the same host **and port** as `registry`. The leak
+/// guard for attaching a token to a tarball download: credentials go to the
+/// registry origin only, never to a third-party CDN a manifest's `dist.tarball`
+/// points at. Port is part of the identity (npm's nerf-dart keys include it),
+/// so a token for `http://localhost:4873` is not sent to `http://localhost:8080`.
 fn same_host(url: &str, registry: &str) -> bool {
-    fn host(s: &str) -> Option<String> {
-        parse_registry(s)?.host_str().map(str::to_ascii_lowercase)
+    fn origin(s: &str) -> Option<(String, Option<u16>)> {
+        let url = parse_registry(s)?;
+        let host = url.host_str()?.to_ascii_lowercase();
+        // `port()` is None for absent or scheme-default ports (the url crate
+        // strips `:443` from https, `:80` from http), so http and https on the
+        // same host still match, while an explicit `:4873` vs `:8080` does not.
+        Some((host, url.port()))
     }
-    matches!((host(url), host(registry)), (Some(a), Some(b)) if a == b)
+    matches!((origin(url), origin(registry)), (Some(a), Some(b)) if a == b)
 }
 
 /// The Bearer token to attach when fetching `url`, or `None` when there is no
@@ -287,9 +304,14 @@ mod tests {
             "https://registry.npmjs.org/pkg/-/pkg-1.0.0.tgz",
             "registry.npmjs.org"
         ));
-        // Bare host with a port (e.g. a local Verdaccio) → matches by host.
+        // Bare host with a port (e.g. a local Verdaccio) → matches by host+port.
         assert!(same_host(
             "http://localhost:4873/pkg/-/pkg-1.0.0.tgz",
+            "localhost:4873"
+        ));
+        // Same host but a DIFFERENT explicit port → never leak the token.
+        assert!(!same_host(
+            "http://localhost:8080/pkg/-/pkg-1.0.0.tgz",
             "localhost:4873"
         ));
         // Host case is normalized.
--- a/crates/pm/src/util/npmrc.rs
+++ b/crates/pm/src/util/npmrc.rs
@@ -1,0 +1,412 @@
+//! Minimal npm `.npmrc` reader (compat layer).
+//!
+//! utoo's native config lives in `.utoo.toml` / `~/.utoo/config.toml`, but the
+//! npm ecosystem configures clients through `.npmrc` files (registry bridges,
+//! private registries, corporate proxies all drop one). Reading them lets `ut`
+//! work in such projects without utoo-specific setup.
+//!
+//! Supported subset:
+//! - `registry=<url>`
+//! - `//host[/path]/:_authToken=<token>` credentials, matched with npm's
+//!   walk-up ("nerf dart") rules
+//! - `${VAR}` environment interpolation in values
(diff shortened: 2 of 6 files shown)


crates/pm/src/util/user_config.rs
```diff
@@ -42,29 +42,59 @@ fn default_cache_dir() -> PathBuf {
...
-        } else {
-            // No config, auto-select fastest registry
-            (select_fastest_registry().await, "auto-select")
+    // init_registry is the process config chokepoint (see main.rs):
+    // `load_levels` populates the merged-config cache that later ConfigValue
+    // reads rely on, and returns the per-level views so the project `.npmrc`
+    // can slot between the project `.utoo.toml` and the user-level global
+    // config (see `RegistrySources` for the precedence rationale).
+    let (levels, npmrc) = tokio::join!(Config::load_levels(), super::npmrc::load());
+    let (global_config, local_config) = levels.unwrap_or_default();
```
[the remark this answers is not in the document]
fengmk2: Good catch, applied in 3003695. `init_registry` now propagates the error with `?` (via `.context("failed to load utoo config")`) instead of `unwrap_or_default()`, so a malformed or unreadable `.utoo.toml` / `~/.utoo/config.toml` fails early with the offending file and reason instead of silently resolving against the wrong registry.

crates/pm/src/util/npmrc.rs
```diff
@@ -0,0 +1,366 @@
...
+    match crate::fs::read_to_string(path).await {
+        Ok(content) => parse(&content),
+        Err(e) => {
+            if e.kind() != std::io::ErrorKind::NotFound {
+                tracing::debug!("skipping unreadable npmrc at {}: {e}", path.display());
+            }
+            HashMap::new()
+        }
+    }
+}
```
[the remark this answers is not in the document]
fengmk2: Kept `.npmrc` reading best-effort here on purpose. Unlike utoo's own config, `.npmrc` is an optional foreign compat layer, and `load()` is memoized in a `OnceCell<Npmrc>` consumed by both `init_registry` and the `resolve_token` path, so making it fallible would ripple through both and let a stale or unrelated `~/.npmrc` line break commands that never touch it. To avoid silently swallowing a real problem, a present-but-unreadable file (permissions, etc.) now logs at `warn` instead of `debug` (3003695); a missing file stays silent since that is the norm. The authoritative-config case is now hard-failing per the user_config.rs thread.

crates/pm/src/util/npmrc.rs
```diff
@@ -0,0 +1,366 @@
...
+/// `//npm.corp/api/npm/repo/:_authToken` wins over it.
+fn auth_token_from(values: &HashMap<String, String>, registry: &str) -> Option<String> {
+    // parse_registry (not raw Url::parse) so scheme-less registry values like
+    // `localhost:4873` match, consistent with the rest of the auth stack.
+    let url = parse_registry(registry)?;
+    let host = url.host_str()?;
+    let mut reg_key = match url.port() {
+        Some(port) => format!("//{host}:{port}{}", url.path()),
+        None => format!("//{host}{}", url.path()),
+    };
```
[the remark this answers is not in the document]
fengmk2: This is already handled by the url crate: `Url::port()` returns `None` for a scheme-default port, so `https://registry.npmjs.org:443` parses with `port() == None` and `reg_key` becomes `//registry.npmjs.org/`, which matches the standard `//registry.npmjs.org/:_authToken` key. Verified against the pinned toolchain (`https://registry.npmjs.org:443` -> `port()=None`, `http://localhost:80` -> `port()=None`, `http://localhost:4873` -> `port()=Some(4873)`), and added a regression test `auth_token_matches_registry_with_explicit_default_port` in 3003695. The suggested branch would be unreachable, since `port()` never returns `Some(443)` / `Some(80)`.

crates/pm/src/service/auth.rs
```diff
@@ -43,9 +47,14 @@ pub async fn resolve_token(registry: &str) -> Option<String> {
...
-    let token = config.get(&token_key(registry)).ok().flatten()?;
-    if token.is_empty() { None } else { Some(token) }
+    if let Ok(config) = Config::load(ConfigScope::Global).await
+        && let Ok(Some(token)) = config.get(&token_key(registry))
+        && !token.is_empty()
+    {
+        return Some(token);
+    }
+
+    crate::util::npmrc::load().await.auth_token_for(registry)
```
elrrrrrrr: 需要参照 code-guard top-use 规则

crates/pm/src/util/user_config.rs
```diff
@@ -42,29 +42,62 @@ fn default_cache_dir() -> PathBuf {
...
-            (reg.clone(), "config")
-        } else {
-            // No config, auto-select fastest registry
-            (select_fastest_registry().await, "auto-select")
+    // init_registry is the process config chokepoint (see main.rs):
+    // `load_levels` populates the merged-config cache that later ConfigValue
+    // reads rely on, and returns the per-level views so the project `.npmrc`
+    // can slot between the project `.utoo.toml` and the user-level global
+    // config (see `RegistrySources` for the precedence rationale).
+    let (levels, npmrc) = tokio::join!(Config::load_levels(), super::npmrc::load());
```
elrrrrrrr: 需要保持 env 和 --registry 参数优先，如果没有再去加载 project 和 user 的 profile

crates/pm/src/service/auth.rs
fengmk2: Done in 3acc60c: added `use crate::util::npmrc;` to the top-level crate imports and changed the call site to `npmrc::load()`, matching the file convention (`Config`/`ConfigScope`/etc. are all top-imported) instead of the inline fully-qualified `crate::util::npmrc::load()` path.

crates/pm/src/util/user_config.rs
fengmk2: Done in 3acc60c. `init_registry` now short-circuits: `--registry`, then `UTOO_REGISTRY`, are checked first and win outright; only when neither is set do we read the profile registries (`.utoo.toml` / project `.npmrc` / global config / user `.npmrc`), and `.npmrc` is loaded only inside that branch. So no profile registry (and no `.npmrc` read) happens when an explicit registry is given.

One thing I deliberately kept unconditional: `Config::load_levels()` (which reads `.utoo.toml` + `~/.utoo/config.toml`). It seeds the shared merged-config cache that every other setting depends on (cache-dir, catalogs, concurrency limits) and must snapshot the invocation cwd before a later workspace/publish `chdir`, so it cannot be gated on the registry inputs. Its `registry` key is simply ignored when `--registry`/env wins. If you would rather that whole load also be deferred, that is a broader change to the config-cache seeding and I can do it as a follow-up.

elrrrrrrr (approved):
LGTM

PR #3235, opened 2026-07-14
fix(pack): ignore unreachable AMD dependencies in Node-target UMD bundles

### LINKED ISSUE #3234: Utoopack resolves unreachable AMD dependencies in Node-target UMD bundles
## Problem

When bundling for Node.js, Utoopack resolves dependencies declared in an unreachable AMD branch of a common UMD wrapper.

This affects legacy but still widely used packages such as `eventproxy`. Its Node.js branch is valid, but the build fails because an AMD-only module does not exist.

## Minimal reproduction

Create `index.js`:

```js
!(function(definition) {
  if (typeof define === 'function') {
    define([ 'missing-amd-only' ], definition);
  } else if (typeof module !== 'undefined' && module.exports) {
    module.exports = definition(require('node:path'));
  }
})(function(path) {
  return path;
});
```

Use this file as an entry and build it with Utoopack's Node.js target (the same target used by `egg-bin bundle`).

## Actual behavior

The build tries to resolve the AMD dependency and fails:

```text
Module not found: Can't resolve 'missing-amd-only'
```

The real-world error from `eventproxy@1.0.0` is:

```text
./node_modules/eventproxy/lib/eventproxy.js:11:5
Module not found: Can't resolve 'eventproxy_debug'
```

`eventproxy@0.3.5` produces the same error.

## Expected behavior

For a Node.js target, `typeof define === 'function'` is false unless AMD `define` is explicitly provided. The AMD branch should be eliminated or ignored, and only the CommonJS branch should be resolved.

The example should bundle successfully and retain the `node:path` dependency without attempting to resolve `missing-amd-only`.

## Context

- Triggered through `egg-bin bundle` / `@eggjs/egg-bundler`
- Utoopack reports the module resolution error
- Updating `eventproxy` does not help because `1.0.0` is the latest release and still contains this UMD wrapper
- Forking or patching every legacy UMD dependency would hide the underlying Node-target dead-branch handling issue

It would also be useful to clarify whether Utoopack intends to support AMD dependency arrays inside UMD wrappers, or whether Node-target builds should always treat an unbound `define` as unavailable.

elrrrrrrr:
## Summary

- mark the free `define` variable as unavailable for Node-target compile-time analysis
- let Turbopack eliminate AMD-first UMD branches before resolving dependency arrays
- add a Node snapshot covering an unreachable missing AMD dependency while retaining `node:path`

## Root cause

Turbopack supports AMD dependency arrays and compile-time free-variable analysis, but Utoopack did not model `typeof define` as `"undefined"` for its Node target. As a result, dependencies in an unreachable AMD branch were still collected and resolved.

## Validation

- `cargo fmt`
- focused `pack-tests` snapshot: 1 passed
- Tombi formatting check
- targeted typos check
- `git diff --check`

Closes #3234


fireairforce (changes requested):
need to update, i will follow up

fireairforce (approved):

PR #3259, opened 2026-07-22
feat(pack): support multiple server entries

fireairforce:
## Summary

- extend `server.entry` to accept either the legacy string form or an array of named server entries
- treat the first array item as the primary server runtime that receives Server Functions, while emitting every item as an independently addressable bundle
- extract dependencies shared by multiple server entries into a Node-loadable shared chunk
- keep server stats entrypoint mappings stable and complete across production builds and dev rebuilds
- include shared chunk paths in entry content hashes so shared dependency updates invalidate every affected entry
- add production snapshots, independent Node execution assertions, and dev rebuild coverage

## Why

The EVJS Utoopack adapter needs to compile one global server runtime together with multiple page-level server entries. Utoopack previously accepted only one server entry, so page server entries could not be emitted or mapped through server stats independently.

Using a single `server.entry` field keeps the configuration aligned with the existing client entry API and avoids exposing the internal primary/additional entry distinction as separate configuration keys.

## Configuration

`server.entry` remains backward compatible with the legacy string form:

```ts
server: {
  entry: "./src/server.ts",
}
```

It now also accepts an array of named entries:

```ts
server: {
  entry: [
    { name: "server", import: "./src/server.ts" },
    { name: "index-server", import: "./src/pages/index.server.ts" },
    { name: "detail-server", import: "./src/pages/detail.server.ts" },
  ],
  output: {
    path: "./dist/server",
    filename: "entries/[name].[contenthash:8].js",
    chunkFilename: "chunks/[name].[contenthash:8].js",
  },
}
```

| Option | Type | Description |
| --- | --- | --- |
| `server.entry` | `string \| Array<{ name: string; import: string }>` | A legacy single entry or an ordered list of named server entries. |
| `server.entry[].name` | `string` | Required unique entry name, used by emitted bundles and server stats. |
| `server.entry[].import` | `string` | Required module path for the server entry. |
| `server.output.path` | `string` | Server output directory relative to the project root; defaults to `{output.path}/server`. |
| `server.output.filename` | `string` | Entry chunk filename template used by the named-entry form. |
| `server.output.chunkFilename` | `string` | Shared/non-entry chunk filename template used by the named-entry form. |

In array form, the first item is the primary server runtime and receives Server Functions. Entry names must be unique. The legacy string form keeps the existing single `index` entry behavior.


hyz-zhao:
#3170

hyz-zhao:
#3170

elrrrrrrr (approved):

PR #3278, opened 2026-08-04
feat(config): make the output order of utoo config list stable

YYYIOU:
Summary
Make the output order of utoo config list stable.
Currently, configuration values are read from a HashMap, which results in a non-deterministic output order. Stabilizing the order improves readability and makes it easier to test and compare outputs.

Changes included:

Sort ordinary config keys alphabetically.

Sort array config keys alphabetically.

Keep the existing output format unchanged.

Add tests to verify order stability.

Test Plan
Run utoo config list multiple times and confirm that the output order is consistent across runs.

Verify that no existing configuration behavior is altered.

Execute the newly added test suite and ensure all tests pass.


fireairforce (approved):

YYYIOU:
#3165

YYYIOU:
#3165

PR #3284, opened 2026-08-06
fix(pm): normalize workspace path filters

gxkl:
## Summary

- normalize workspace path filters before matching
- accept explicit relative paths such as `./tools/egg-bin`
- normalize Windows separators and trailing slashes
- add unit and CLI contract regression coverage

## Motivation

The npm-installed Windows PowerShell launcher can pass `--workspace ./tools/egg-bin` directly to utoo when the caller uses `--` for argument forwarding. Workspace discovery stores the path without the leading `./`, so the literal filter previously failed to select the package.

Normalizing both discovered workspace paths and CLI path filters makes these equivalent forms match consistently across platforms.

## Tests

- `cargo fmt --all --check`
- workspace service tests: 9 passed
- `cargo test -p utoo-pm --test cli_contract`: 29 passed

### DIFF
--- a/crates/pm/src/service/workspace.rs
+++ b/crates/pm/src/service/workspace.rs
@@ -244,18 +244,26 @@ pub struct WorkspaceJson {
 /// Each filter is matched against the workspace name and its project-relative
 /// path via [`matches_pattern`], which handles both literal equality and the
 /// `*` glob forms supported elsewhere in the codebase (e.g. `packages/*`).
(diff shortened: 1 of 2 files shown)


gxkl:
Closing this approach because the workspace-path normalization happens after PowerShell has already stripped the `--` delimiter. It makes the Windows command execute a workspace package script instead of preserving the intended arguments passed to the root tsdown script. The fix should be made at the workflow or Windows launcher boundary instead.

PR #3289, opened 2026-08-10
fix(pm): reject duplicate workspace package names

weiye1234562:
## Summary

- detect duplicate workspace package names during shared workspace discovery, before dependency graph or topology construction
- return a stable `EDUPLICATEWORKSPACE` error containing the conflicting package name and sorted `package.json` paths
- add Ruborist and PM regression coverage
- enable the Arborist `workspaces-duplicate` fixture and remove its skip documentation
- clean up Windows-only lint noise uncovered by the required all-targets Clippy check

## Root cause

Workspace members are indexed by package name. When multiple workspace manifests declared the same `name`, later entries could overwrite earlier entries in name-keyed maps, making dependency resolution and topology selection ambiguous.

## User impact

Invalid workspace projects now fail early with an actionable error, while valid workspace discovery and topology behavior remain unchanged.

## Validation

- `cargo fmt`
- `cargo test -p utoo-ruborist resolver::workspace` (4 passed)
- `cargo test -p utoo-pm service::workspace::tests` (10 passed)
- `./e2e/pm-arborist.sh workspaces-duplicate` (1 passed, 0 failed, 0 skipped)


elrrrrrrr (approved):
LGTM

elrrrrrrr (approved):

PR #3296, opened 2026-08-11
feat(pack): support server-specific resolve options

fireairforce:
## Summary

- add `server.resolve` for server-only module resolution configuration
- merge `server.resolve.alias` over shared `resolve.alias` entries so server bundles can target different modules
- let `server.resolve.extensions` replace shared extensions when configured, while falling back to `resolve.extensions` when omitted
- update TypeScript types and the generated JSON schema
- add snapshot coverage proving client and server builds resolve different aliases and extensions

## Test Plan

- `cargo fmt --check`
- `cargo test -p pack-schema`
- `cargo test -p pack-tests --test snapshot test_tests__snapshot__basic__server_alias -- --exact`
- `biome check packages/pack-shared/src/config.ts`


PeachScript (approved):
+1

PR #3299, opened 2026-08-12
feat(pm): support packageManager self-pinning

fireairforce:
## Summary

Projects can now pin an exact Utoo package-manager version with `"packageManager": "utoo@x.y.z"`, so contributors and CI use the same implementation instead of whichever global Utoo version happens to be installed.

- Honor exact `packageManager` pins for local dependency operations.
- Provision the matching supported platform package from the configured npm registry.
- Verify registry SRI/shasum metadata, cache the release in a platform-specific slot under a dedicated self-pin cache root, and revalidate the cached executable and reported version before handoff.
- Treat invalid cached releases as cache misses: never execute them, and replace only the affected platform slot after a newly downloaded archive passes registry integrity verification.
- Preserve command arguments, working directory, stdio, exit status, and signals during handoff.
- Keep scripts, `ut x`, query commands, and global installs on the current executable.
- Document the pinning flow and the `UTOO_SELF_PIN=0` escape hatch.

Prior art: Nub provides the same project-level version selection through [`packageManager: "nub@x.y.z"`](https://nubjs.com/docs/pm#pinning-nub-itself), provisioning and handing off to the exact pinned Nub release during install operations.

SchemaStore support is being added in [SchemaStore/schemastore#6192](https://github.com/SchemaStore/schemastore/pull/6192). It updates the `package.json` schema pattern from:

```diff
-(npm|pnpm|yarn|bun|aube|nub)@\d+\.\d+\.\d+(-.+)?
+(npm|pnpm|yarn|bun|aube|nub|utoo)@\d+\.\d+\.\d+(-.+)?
```

The first release containing this feature should be `utoo@1.1.8` or newer; older binaries do not understand the recursion/auto-update guard used by the handoff.

## Test Plan

- `cargo test -p utoo-pm` — 369 unit tests and 28 CLI contract tests passed; 3 network tests ignored by the existing suite.
- Added regression coverage for platform-specific cache keys, malformed cache metadata, executable integrity mismatches, and self-pin cleanup serialization.
- Keep the self-pin cache physically isolated from ordinary registry packages while exposing logical `_utoo-self-<platform>@<version>` entries for exact `utoo clean` matching. Cleanup acquires the same platform/version lock as handoff and provisioning.
- On Windows, keep the slot locked through child process creation, then release it before waiting so nested `utoo clean` invocations cannot deadlock the handoff parent.
- `cargo clippy --all-targets -- -D warnings --no-deps`.
- `cargo fmt --all -- --check`.
- `npx tombi format --check`.
- `typos` on changed files.
- Manual corruption checks confirmed modified cached releases are never executed and fail closed if verified reprovisioning is unavailable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,7 +80,7 @@ uses the same package-manager implementation:
 
 When a local dependency command runs with another Utoo version, Utoo downloads
 the matching native release, verifies the registry checksum, caches it under
-`~/.cache/nm/self/<version>@<platform>`, and hands the command off with the same
+`~/.cache/nm/self-<platform>/<version>`, and hands the command off with the same
 arguments, working directory, and exit behavior. Warm caches work offline. An
 invalid cached release is never executed and is reprovisioned from the registry
 when available. This applies to local install/add, uninstall, update, rebuild,
--- a/README.md
+++ b/README.md
@@ -69,6 +69,23 @@ ut add lodash       # Add a package (or use `ut add`)
 ut x create-react   # Execute a package (npx style, or use `ut x`)
 ```
 
+Pin the project to an exact Utoo release so every contributor and CI runner
+uses the same package-manager implementation:
+
+```json
+{
+  "packageManager": "utoo@1.1.8"
+}
+```
+
+When a local dependency command runs with another Utoo version, Utoo downloads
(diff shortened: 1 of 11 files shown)


elrrrrrrr (approved):
LGTM

PR #3346, opened 2026-09-10
fix(pack): honor targets in library runtime output

fireairforce:
Library builds configured for older browsers could still contain optional chaining in automatic public-path detection and modern syntax in module factories and runtime wrappers. The library chunking context replaced the requested environment with an EdgeWorker/Node 18 environment to keep dynamic imports in one file.

Preserve the platform's actual environment and use `ChunkLoading::SingleChunk` to retain single-file library output. Apply preset-env to the assembled library before minification so runtime assets and generated async-module wrappers also obey the target. Inline transform helpers stay inside the library, source maps are composed, and minification cannot reintroduce unsupported arrow functions. Re-emit minified legacy output as ES5 to preserve required do/while semicolons; ES5 emission also normalizes binary numeric literals. This final emission does not run transforms or inject helpers again. Bootstrap `globalThis` where the target requires it.

When preset-env leaves the resolved AST unchanged, return the original code and sectioned source map directly. This preserves non-minified layout, comments, line positions, and source-map structure for targets that need no lowering. Legacy ES5 output still goes through emission to normalize raw literals and statement separators. All existing snapshots remain unchanged. Add one readable, unminified legacy runtime snapshot in the existing runtime group, with an ES5 parse and execution assertion covering auto public-path fallback, inline dynamic imports, helper scope, and missing globalThis. The example-level temporary validation scripts are not included.

Validation:
- Negative control: the original library runtime snapshot fails Acorn's ES5 parse at its outer runtime arrow.
- Six temporary local integration checks passed: legacy production/minified/development, modern minified/unminified, and Chrome 70. They verify ES5/ES2019 parsing, original formatting for compatible targets, browser/CommonJS exports, dynamic imports, public-path fallback, helper scope, source maps, and license extraction. Chrome 70 still lowers optional chaining despite supporting arrow functions. The validation scripts were removed after use.
- Snapshot suite: 147 passed, including the new legacy runtime case and all unchanged snapshots from next (2b393d02e).
- A downstream widget library: both minified and unminified UMD pass ES5 and ES2019 parsing, with zero optional chains and arrow functions.
- `cargo fmt --check`
- `npx biome ci`, `npx tombi format --check`, `typos`

This fixes generated JavaScript syntax. General runtime API polyfills remain a separate concern.


xusd320 (approved):

PeachScript (approved):
+1

